PR #3, opened 2019-01-11
Deposit

bh2smith:
Include basic deposit functionality with units tests.

**Note that** deposit hashed are kept as an array for the purpose of implementing rollbacks in the future.


contracts/BatchAuction.sol
```diff
@@ -38,4 +54,32 @@ contract BatchAuction is Ownable {
...
+
+        require(
+            ERC20(tokenAddress).transferFrom(msg.sender, address(this), amount), 
+            "Unsuccessful transfer"
+        );
+        
+        // Increment depositIndex until it matches correct.
+        uint nextDepositIndex = block.number / 20;
+        while (depositIndex != nextDepositIndex) {
+            depositIndex++;
```
josojo: please use here depositIndex = block.number/20

If there were for 60 blocks no deposits, I would prefer to not do the next 3 deposits split over 3 snarks

josojo (approved):
very good PR. Thanks.

After fixing the one raised comment, you are ready to merge

contracts/BatchAuction.sol
bh2smith: Done! Great suggestion!

fleupold (commented):
Sorry I wasn't able to review this change earlier. I think there is a fundamental problem with the way deposits are modeled here if we want to be able to prove them via snarks.

Currently, the deposit hash is bound in time (20 blocks) but unbound in size (how many deposits are "inside" it). However, the snark to apply any state transition has to take a fixed size input. While it's possible to collect less than that limit (we can pad the remaining values with 0), it won't be possible to ever process more deposits (268M R1CS size limit).
Thus I believe we need to limit the maximum number of deposits per hash.

Would you be able to explain, why you chose to bound each deposit state by time?

bh2smith:
@Felix the Block time model for deposits is meant to ensure deposits occur in batches and that no-inflight deposits could affect the deposit hash while it is trying to be submitted.

Issue #7, opened 2019-01-17
Include slotIndex (as counter within each slot)


bh2smith:
This will be necessary for the _driver_ to know in which order the deposits are to be applied. Hence, this index will also need to be emitted with the deposit event.

PR #8, opened 2019-01-18
Include slot index on current implementation of deposits

bh2smith:
This is a reasonable temporary solution to #6


josojo (approved):
Great!

bh2smith:
closes #7

PR #10, opened 2019-01-19
#9 remove depositHash check in apply deposit

### LINKED ISSUE #9: Remove depositHash verification in applyDeposit
Because the contract verifies that the deposit slot is inactive, the additional layer of security which ensures the sender is applying the correct deposits is unnecessary

bh2smith:
Closes #9


josojo (approved):
Thanks man!

to be honest, firstly, I thought we would only remove the depositHash from the logging. But yeah, you are right, also over here it is not needed.

PR #13, opened 2019-01-21
Setup scripts and contract interaction

josojo:
script can be run with:  truffle exec scripts/setup_initial_deposit.js 

Needs to handle various error messages. I will not do it for now.


bh2smith (approved):
This thing is great! Works very well for deposits and setup environment!

scripts/deposit.js
```diff
@@ -0,0 +1,24 @@
...
+    const instance = await SnappBase.deployed()
+    const [accountId, tokenId, amount] = await process.argv.slice(4)
+
+    const depositor = await instance.accountToPublicKeyMap.call(accountId)
+    if (depositor == zero_address) {
+        console.log("No account registerd at index %s", accountId)
+        callback()
+    }
+    const slot = (await instance.depositSlot.call()).toNumber()
+    const tx = await instance.deposit(tokenId, amount, {from: depositor})
```
josojo, line 31: It would be better to get the slot number from the event. The way you have done it, it will only be correct in 99 percent of all cases on a public testnet

josojo, line 31: You might also wanna check that the user has the funds available and throw overwise..

Where do you approve the token transfer? IF it is only approved in setup_environment, you might need to say this at the beginning of the file. I would recommend doing a seperate approval, so that the deposit script can be used independently

bh2smith: Yes, but these scripts aren't meant to listen to events... Oh! You mean I can get the event from the transaction?!

bh2smith: The approval is done in SetupEnvironment.

bh2smith: I can separate them and make three scripts:

- setup
- approve
- setup-and-approve

bh2smith: Yes I will absolutely do this.

bh2smith: I don't fully understand what you mean by deposit script being used independently... If we don't setup environment, we will also need a register token, open account, and a bunch of stuff.... I guess we will need these anyways. I really need to figure out how to import files.

josojo: Okay, it would be for later for sure. But I could come up with some use case. E.g. a bot has already an account and he just wants to deposit some tokens ....

bh2smith: Done!

bh2smith: I have included pre-validation in the deposit request!

PR #21, opened 2019-01-24
new getters

josojo:
I need these getters for my driver

Additional to the getter functions I "uncommented" the networksettings for development. This makes the truffle setup easier on my machine. Is this okay? Why don't we have any truffle settings? Or is this just a backup setting file and I should have my config in truffle.js?


bh2smith (approved):
I wasn’t aware that public view could be called without this.

I really thought the development network was uncommented already! This is all good for me!

PR #23, opened 2019-01-24
new smartcontract getter function

josojo:
tiny pr needed from my side


bh2smith (approved):
Nice and easy to approve! At some point we should include Doc Strings to our code and reorder the different types into appropriate categories

PR #24, opened 2019-01-24
Complete rework of deposits

### LINKED ISSUE #5: Deposit Index Gap
Having deposit slots determined by the integer division `blockNumber / 20` creates massive gaps in the deposit index. For example, (as of this date) the first non-empty deposit index would be `7064235 / 20 = 353211`. Furthermore, whenever there is a lull in deposits, there will be empty deposit blocks between different indices.

This can be avoided by using a depostHashes array (rather than mapping) which is continually appended to at the appropriate time (rather block time or deposit limit reached).

bh2smith:
Takes care of slot limit and removes gaps. Only empty slot is possibly the first

Closes #5 and #6

### DIFF
--- a/contracts/SnappBase.sol
+++ b/contracts/SnappBase.sol
@@ -6,8 +6,9 @@ import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 
 contract SnappBase is Ownable {
 
-    uint16 public constant MAX_ACCOUNT_ID = 100;
+    uint16 public constant MAX_ACCOUNT_ID = 100;     // TODO - make larger or use uint8
     uint8 public constant MAX_TOKENS = 30;
+    uint16 public constant MAX_DEPOSIT_BATCH = 100;  // TODO - make larger or use uint8
 
     bytes32[] public stateRoots;  // Pedersen Hash
 
@@ -25,12 +26,13 @@ contract SnappBase is Ownable {
     mapping (uint8 => address) public tokenIdToAddressMap;
 
     struct DepositState {
-        bytes32 shaHash;
-        bool applied;
+        uint16 size;                   // Number of deposits in this batch
+        bytes32 shaHash;               // Rolling shaHash of all deposits
+        uint creationBlock;            // Timestamp of batch creation
+        uint appliedAccountStateIndex; // accountState index when batch applied (for rollback)
     }
 
-    uint16 public slotIndex;
-    mapping (uint => DepositState) public depositHashes;
+    DepositState[] deposits;
 
     event Deposit(uint16 accountId, uint8 tokenId, uint amount, uint slot, uint16 slotIndex);
     event StateTransition(TransitionType transitionType, uint from, bytes32 to, uint slot);
@@ -45,6 +47,16 @@ contract SnappBase is Ownable {
         // The initial state should be Pederson hash of an empty balance tree
         bytes32 stateInit = bytes32(0);  // TODO
         stateRoots.push(stateInit);
+        deposits.push(
+            DepositState({
+                size: 0,
+                shaHash: bytes32(0),
+                creationBlock: block.number,
+                appliedAccountStateIndex: 0
+            })
+        );
+
+
         emit SnappInitialization(stateInit, MAX_TOKENS, MAX_ACCOUNT_ID);
     }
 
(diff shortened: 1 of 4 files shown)


contracts/SnappBase.sol
```diff
@@ -25,12 +26,13 @@ contract SnappBase is Ownable {
...
-        bytes32 shaHash;
-        bool applied;
+        uint16 size;                   // Number of deposits in this batch
+        bytes32 shaHash;               // Rolling shaHash of all deposits
+        uint creationBlock;            // Timestamp of batch creation
+        uint appliedAccountStateIndex; // accountState index when batch applied (for rollback)
     }
 
-    uint16 public slotIndex;
-    mapping (uint => DepositState) public depositHashes;
```
josojo, line 33: I would always prefer to work with mappings in such a situation. Especially, since we want to go later to modular representation and erasing old slots.

josojo (approved):
Wow man! 
thanks for this piece of cake! you are soo quick

I have one concern, but I can not come up with a better method. If there are very rarely any deposits, then no one might be triggering the next slot. And then the operator can not credit the deposits.

With the current spec, the operator would have to make a deposit and then apply them. 

We could make it more complicated by allowing the operator to push into new slots directly. But this would increase the complexity. That is the reason why I am against it. But it is not good for the user experience, if deposits are not credited regularly...

contracts/SnappBase.sol
bh2smith: Yes I also like mapping. What should I do now? Try to do the same stuff with a mapping?

josojo: Yes, I would do the same with the mapping.

contracts/SnappBase.sol
```diff
@@ -109,23 +133,30 @@ contract SnappBase is Ownable {
...
         public onlyOwner()
     {   
-        require(slot < depositSlot(), "Deposit slot must exist and be inactive");
-        require(depositHashes[slot].applied == false, "Deposits already processed");
-        require(depositHashes[slot].shaHash != bytes32(0), "Deposit slot is empty");
+        require(slot <= depositIndex(), "Requested deposit slot does not exist");
+
+        require(slot == 0 || deposits[slot-1].appliedAccountStateIndex != 0, "Must apply deposit slots in order!");
+        require(deposits[slot].appliedAccountStateIndex == 0, "Deposits already processed");
+        require(block.number > deposits[slot].creationBlock + 20, "Requested deposit slot is still active");
```
bh2smith, line 129: @josojo This line here says that any deposit slot can be applied, even if nobody has deposited for a long time. Notice the same comparison with the block time + 20. This means that deposit slots become in active after 20 blocks, even if nobody deposits in the next one.

contracts/SnappBase.sol
```diff
@@ -109,23 +133,30 @@ contract SnappBase is Ownable {
     )
         public onlyOwner()
     {   
-        require(slot < depositSlot(), "Deposit slot must exist and be inactive");
-        require(depositHashes[slot].applied == false, "Deposits already processed");
-        require(depositHashes[slot].shaHash != bytes32(0), "Deposit slot is empty");
+        require(slot <= depositIndex(), "Requested deposit slot does not exist");
```
bh2smith: This requirement only makes sure that the index specified isn't out of range.

bh2smith:
@joso, your concern in the previous comment is already considered and taken care of. Please see my inline comment at line 140.

contracts/SnappBase.sol
```diff
@@ -80,22 +92,34 @@ contract SnappBase is Ownable {
...
+        DepositState memory currDepositState = deposits[depositIndex()];
+        if (currDepositState.size == MAX_DEPOSIT_BATCH || block.number > currDepositState.creationBlock + 20) {
+            currDepositState = DepositState({
+                size: 0,
+                shaHash: bytes32(0),
+                creationBlock: block.number,
+                appliedAccountStateIndex: 0   // Default 0 implies not applied.
+            });
+            deposits.push(currDepositState);
+            // depositIndex++;
```
josojo: you can remove this one

contracts/SnappBase.sol
```diff
@@ -32,7 +32,9 @@ contract SnappBase is Ownable {
         uint appliedAccountStateIndex; // accountState index when batch applied (for rollback)
     }
 
-    DepositState[] public deposits;
+    uint public depositIndex=0;
```
bh2smith: Default value of an integer is 0 already.

contracts/SnappBase.sol
```diff
@@ -25,12 +26,15 @@ contract SnappBase is Ownable {
...
-        bool applied;
+        uint16 size;                   // Number of deposits in this batch
+        bytes32 shaHash;               // Rolling shaHash of all deposits
+        uint creationBlock;            // Timestamp of batch creation
+        uint appliedAccountStateIndex; // accountState index when batch applied (for rollback)
     }
 
-    uint16 public slotIndex;
-    mapping (uint => DepositState) public depositHashes;
+    uint public depositIndex=0;
```
bh2smith:
```suggestion
    uint public depositIndex;
```

contracts/SnappBase.sol
```diff
@@ -80,22 +91,30 @@ contract SnappBase is Ownable {
...
+            abi.encodePacked(currDepositState.shaHash, accountId, tokenIndex, amount)
         );
-        depositHashes[depositSlot] = DepositState({shaHash: nextDepositHash, applied: false});
+        currDepositState.shaHash = nextDepositHash;
+        currDepositState.size++;
 
-        emit Deposit(accountId, tokenIndex, amount, depositSlot, slotIndex);
-        slotIndex++;
-    }
+        deposits[depositIndex] = currDepositState;
```
bh2smith: Is this actually necessary? I know it was for the array... but might not be for the mapping (since we were mutating what the mapping was already pointing at).

contracts/SnappBase.sol
```diff
@@ -25,12 +26,15 @@ contract SnappBase is Ownable {
...
+        bytes32 shaHash;               // Rolling shaHash of all deposits
+        uint creationBlock;            // Timestamp of batch creation
+        uint appliedAccountStateIndex; // accountState index when batch applied (for rollback)
     }
 
-    uint16 public slotIndex;
-    mapping (uint => DepositState) public depositHashes;
+    uint public depositIndex=0;
+    mapping (uint => DepositState) public deposits;
+    //DepositState[] public deposits;
```
bh2smith: can we remove this now?

contracts/SnappBase.sol
```diff
@@ -80,22 +91,30 @@ contract SnappBase is Ownable {
             "Unsuccessful transfer"
         );
 
-        uint depositSlot = depositSlot();
-        if (depositHashes[depositSlot].shaHash == bytes32(0)) {
-            slotIndex = 0;
+        // uint depositIndex = deposits.length - 1;
```
bh2smith: remove!

PR #42, opened 2019-02-18
Apply Withdraw Script

bh2smith:
This is version is currently working, but the bitmap requirement as an argument is going to be annoying. Some ways that might make it easier is to pass one of the following flags

- "FullBitMap" - an array of all `true`
- "NullBitMap" - an array of all `false`
- "ExclusionIndices" - a list of indices for which the withdraw was excluded
- "InclusionIndices" - a list of indices for which the withdraw was included

Currently this can be tested as follows, 

```
npm run apply-withdrawals <slot: int> <bitmap: bool[100]> <merkleRoot: bytes32> <newState: bytes32>

I still want to include some assertions on the contents of the bitmap here. However, it is working in its current state.

### DIFF
--- a/scripts/apply_withdrawals.js
+++ b/scripts/apply_withdrawals.js
@@ -1,0 +1,28 @@
+const SnappBase = artifacts.require("SnappBase")
+
+module.exports = async (callback) => {
+  try {
+    const arguments = await process.argv.slice(4)
+    if (arguments.length != 4) {
+      callback("Error: This script requires arguments - <slot> <inclusion bitMap> <merkle root> <new state root>")
+    }
+    const [slot, bitmap, merkle_root, new_state] = arguments
+    
+    const instance = await SnappBase.deployed()
+    const state_index = (await instance.stateIndex.call()).toNumber()
+    const curr_state = await instance.stateRoots.call(state_index)
+
+    const withdraw_state = await instance.pendingWithdraws.call(slot)
+    if (withdraw_state.appliedAccountStateIndex != 0) {
+      callback("Error: Requested withdraw slot has already been applied")
+    }
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by josojo** [comment]
If the bitmap is getting too annoying, I would suggest to take it out! 

It's not really needed anyway, as Felix pointed out.

File: scripts/apply_withdrawals.js [hunk 1]
```diff
@@ -0,0 +1,32 @@
+const SnappBase = artifacts.require("SnappBase")
+
+// TODO - may adjust inclusion bitmap to expect shorter input
+// const { falseArray }  = require("../test/snapp_utils.js")
+
+module.exports = async (callback) => {
+  try {
+    const arguments = await process.argv.slice(4)
+    if (arguments.length != 4) {
```
josojo, line 7: Can we do the network parameter slice again?

josojo (approved):

fleupold (approved):
Looks good. For better readability of this PR it would help if it was not based off master, but off the request_withdraw branch instead. Otherwise the changes of both PRs show up in this one.

Who is going to use this script? Is it just for your testing? I was assuming the driver would handle the state transition directly, without using a js script.

bh2smith:
It was originally based off the other branch, but I rebased on master when @josojo mentioned to use the network parameter. Can rebase again before merge. Anyway, I only included this for completeness and it is getting me ready for the "claimWithdraw" script. I also figure, something like this could eventually have a use for someone else in the front end. In case someone wants to drive their own contract.

bh2smith:
I will merge this for now and make a separate PR with the relevant adjustments when removing the bitMap from the contracts function parameters.

PR #58, opened 2019-03-27
#54 Order Collection

bh2smith:
Introduce new contract `SnappAuction` with minimum required attributes to place a limit sell order. 
Include units tests. Will keep the scripting in a separate PR.

### DIFF
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -8,7 +8,7 @@ contract SnappAuction is SnappBase {
     uint16 public constant AUCTION_BATCH_SIZE = 1000;
 
     uint public auctionIndex;
-    mapping (uint => PendingFlux) public auctions;
+    mapping (uint => PendingBatch) public auctions;
 
     event SellOrder(
         uint auctionId, 
@@ -32,20 +32,16 @@ contract SnappAuction is SnappBase {
     ) public onlyRegistered() {
         require(buyToken != sellToken, "Buy and Sell tokens must differ!");
 
-        // Must have 0 < tokenId < MAX_TOKENS anyway, so might as well ensure registered.
+        // Must have 0 < tokenId < MAX_TOKENS anyway, so may as well ensure registered.
         require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
         require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
-
-        // Restrict buy and sell amount to occupy less than 100 bits.
-        require(minBuyAmount < 0x10000000000000000000000000, "Buy amount too large!");
-        require(maxSellAmount < 0x10000000000000000000000000, "Sell amount too large!");
 
         if (
             auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
             block.number > auctions[auctionIndex].creationBlock + 20
         ) {
             auctionIndex++;
-            auctions[auctionIndex] = PendingFlux({
+            auctions[auctionIndex] = PendingBatch({
                 size: 0,
                 shaHash: bytes32(0),
                 creationBlock: block.number,
@@ -75,6 +71,10 @@ contract SnappAuction is SnappBase {
         uint128 buyAmount, 
         uint128 sellAmount
     ) internal pure returns (bytes32) {
+        // Restrict buy and sell amount to occupy at most 100 bits.
+        require(buyAmount < 0x10000000000000000000000000, "Buy amount too large!");
+        require(sellAmount < 0x10000000000000000000000000, "Sell amount too large!");
+
         // solhint-disable-next-line max-line-length
         return bytes32(uint(accountId) + (uint(buyToken) << 16) + (uint(sellToken) << 24) + (uint(sellAmount) << 32) + (uint(buyAmount) << 132));
     }
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -41,8 +41,8 @@ contract SnappAuction is SnappBase {
             block.number > auctions[auctionIndex].creationBlock + 20
         ) {
             require(
-                auctionIndex < 2 ||
-                auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions"
+                auctionIndex < 2 || auctions[auctionIndex - 2].appliedAccountStateIndex != 0,
+                "Too many pending auctions"
             );
             auctionIndex++;
             auctions[auctionIndex] = PendingBatch({
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -27,8 +27,8 @@ contract SnappAuction is SnappBase {
     function placeSellOrder(
         uint8 buyToken,
         uint8 sellToken,
-        uint128 minBuyAmount,
-        uint128 maxSellAmount
+        uint128 buyAmount,
+        uint128 sellAmount
     ) public onlyRegistered() {
         // Must have 0 < tokenId < MAX_TOKENS anyway, so may as well ensure registered.
         require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
@@ -57,12 +57,12 @@ contract SnappAuction is SnappBase {
         bytes32 nextAuctionHash = sha256(
             abi.encodePacked(
                 auctions[auctionIndex].shaHash,
-                encodeOrder(accountId, buyToken, sellToken, minBuyAmount, maxSellAmount)
+                encodeOrder(accountId, buyToken, sellToken, buyAmount, sellAmount)
             )
         );
         auctions[auctionIndex].shaHash = nextAuctionHash;
 
-        emit SellOrder(auctionIndex, auctions[auctionIndex].size, accountId, buyToken, sellToken, minBuyAmount, maxSellAmount);
+        emit SellOrder(auctionIndex, auctions[auctionIndex].size, accountId, buyToken, sellToken, buyAmount, sellAmount);
         // Only increment size after event (so it is emitted as an index)
         auctions[auctionIndex].size++;
     }
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -40,9 +40,10 @@ contract SnappAuction is SnappBase {
             auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
             block.number > auctions[auctionIndex].creationBlock + 20
         ) {
-            if (auctionIndex > 1) {
-                require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions");
-            }
+            require(
+                auctionIndex < 2 ||
+                auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions"
+            );
             auctionIndex++;
             auctions[auctionIndex] = PendingBatch({
                 size: 0,
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -30,11 +30,11 @@ contract SnappAuction is SnappBase {
         uint128 minBuyAmount,
         uint128 maxSellAmount
     ) public onlyRegistered() {
-        require(buyToken != sellToken, "Buy and Sell tokens must differ!");
-
         // Must have 0 < tokenId < MAX_TOKENS anyway, so may as well ensure registered.
         require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
         require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
+
+        // Could also enforce that buyToken != sellToken, but not technically illegal.
 
         if (
             auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -30,10 +30,6 @@ contract SnappAuction is SnappBase {
         uint128 minBuyAmount,
         uint128 maxSellAmount
     ) public onlyRegistered() {
-        if (auctionIndex > 1) {
-            require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "To many pending auctions");
-        }
-
         require(buyToken != sellToken, "Buy and Sell tokens must differ!");
 
         // Must have 0 < tokenId < MAX_TOKENS anyway, so may as well ensure registered.
@@ -44,6 +40,9 @@ contract SnappAuction is SnappBase {
             auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
             block.number > auctions[auctionIndex].creationBlock + 20
         ) {
+            if (auctionIndex > 1) {
+                require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions");
+            }
             auctionIndex++;
             auctions[auctionIndex] = PendingBatch({
                 size: 0,
--- a/contracts/SnappAuction.sol
+++ b/contracts/SnappAuction.sol
@@ -1,0 +1,81 @@
+pragma solidity ^0.5.0;
+
+import "./SnappBase.sol";
+
+
+contract SnappAuction is SnappBase {
+  
+    uint16 public constant AUCTION_BATCH_SIZE = 1000;
+
+    uint public auctionIndex;
+    mapping (uint => PendingFlux) public auctions;
+
+    event SellOrder(
+        uint auctionId, 
+        uint16 slotIndex,
+        uint16 accountId, 
+        uint8 buyToken, 
+        uint8 sellToken, 
+        uint128 buyAmount,
+        uint128 sellAmount
+    );
+
+    constructor () public {
+        auctions[auctionIndex].creationBlock = block.number;
+    }
+
+    function placeSellOrder(
+        uint8 buyToken,
+        uint8 sellToken,
+        uint128 minBuyAmount,
+        uint128 maxSellAmount
+    ) public onlyRegistered() {
+        require(buyToken != sellToken, "Buy and Sell tokens must differ!");
+
+        // Must have 0 < tokenId < MAX_TOKENS anyway, so might as well ensure registered.
+        require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
+        require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
+
+        // Restrict buy and sell amount to occupy less than 100 bits.
+        require(minBuyAmount < 0x10000000000000000000000000, "Buy amount too large!");
+        require(maxSellAmount < 0x10000000000000000000000000, "Sell amount too large!");
+
+        if (
+            auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
+            block.number > auctions[auctionIndex].creationBlock + 20
+        ) {
+            auctionIndex++;
+            auctions[auctionIndex] = PendingFlux({
+                size: 0,
+                shaHash: bytes32(0),
+                creationBlock: block.number,
+                appliedAccountStateIndex: 0
+            });
+        }
+
+        // Update Auction Hash based on request
+        uint16 accountId = publicKeyToAccountMap[msg.sender];
+        bytes32 nextAuctionHash = sha256(
+            abi.encodePacked(
+                auctions[auctionIndex].shaHash,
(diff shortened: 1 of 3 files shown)


contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+        uint128 minBuyAmount,
+        uint128 maxSellAmount
+    ) public onlyRegistered() {
+        require(buyToken != sellToken, "Buy and Sell tokens must differ!");
+
+        // Must have 0 < tokenId < MAX_TOKENS anyway, so might as well ensure registered.
+        require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
+        require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
+
+        // Restrict buy and sell amount to occupy less than 100 bits.
```
fleupold: nit: I believe this should say `to occupy at most 100 bits`

```python
>>> int('1'*100, 2) < 0x10000000000000000000000000
True
```

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+        require(minBuyAmount < 0x10000000000000000000000000, "Buy amount too large!");
+        require(maxSellAmount < 0x10000000000000000000000000, "Sell amount too large!");
+
+        if (
+            auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
+            block.number > auctions[auctionIndex].creationBlock + 20
+        ) {
+            auctionIndex++;
+            auctions[auctionIndex] = PendingFlux({
```
fleupold: I'm not a big fan of reusing PendingFlux here - since a flux is a deposit or withdraw. If the structure has the same data, maybe we can rename PendingFlux to PendingBatch?

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+        auctions[auctionIndex].creationBlock = block.number;
+    }
+
+    function placeSellOrder(
+        uint8 buyToken,
+        uint8 sellToken,
+        uint128 minBuyAmount,
+        uint128 maxSellAmount
+    ) public onlyRegistered() {
+        require(buyToken != sellToken, "Buy and Sell tokens must differ!");
```
fleupold: Interesting. This might be somewhat related to all the requires we found a bit difficult to work with in the dutchX when we tried to build a smart contract that wants to use it.

I agree that it doesn't make sense to place an order in the same currency (just like it doesn't make sense to withdraw 0 tokens), but maybe we should only throw when someone is trying to do something illegal.

Don't need to change, but something to think about/discuss on a more philosophical level.

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    function encodeOrder(
+        uint16 accountId, 
+        uint8 buyToken, 
+        uint8 sellToken, 
+        uint128 buyAmount, 
+        uint128 sellAmount
+    ) internal pure returns (bytes32) {
+        // solhint-disable-next-line max-line-length
```
fleupold, line 82: Maybe comment that it's critical to check buyAmount and sellAmount are at most 100 bit which happens above?

I'm somewhat afraid we might remove this require at some point and not update this code. Should we maybe trim any bits > 100?

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+        if (
+            auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
+            block.number > auctions[auctionIndex].creationBlock + 20
+        ) {
+            auctionIndex++;
+            auctions[auctionIndex] = PendingFlux({
+                size: 0,
+                shaHash: bytes32(0),
+                creationBlock: block.number,
+                appliedAccountStateIndex: 0
```
fleupold, line 52: Is this relevant for orders? Will it be the state index in which they are processed?

test/snapp_auction.js
```diff
@@ -0,0 +1,97 @@
...
+      await setupEnvironment(MintableERC20, instance, token_owner, [user_1], 2)
+      await instance.placeSellOrder(1, 2, 1, 1, { from: user_1 })
+
+      await waitForNBlocks(21, owner)
+      await instance.placeSellOrder(1, 2, 1, 1, { from: user_1 })
+
+      const auctionIndex = (await instance.auctionIndex.call()).toNumber()
+      assert.equal(auctionIndex, 1)
+    })
+  })
```
fleupold, line 104: I believe in order to get 100% branch coverage, we'd have to test the case where we fill a batch just by pushing a lot of orders in the same block.

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+        require(tokenIdToAddressMap[buyToken] != address(0), "Buy token is not registered");
+        require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
+
+        // Restrict buy and sell amount to occupy less than 100 bits.
+        require(minBuyAmount < 0x10000000000000000000000000, "Buy amount too large!");
+        require(maxSellAmount < 0x10000000000000000000000000, "Sell amount too large!");
+
+        if (
+            auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
+            block.number > auctions[auctionIndex].creationBlock + 20
```
fleupold, line 41: I'm a bit concerned about this condition. This could lead to a devil spiral if our processing time takes longer than 20 blocks. Then we would create new batches faster then we process them leading to an evergrowing backlog.

I wonder if it would make sense to somehow limit new batch creation to a maximum of previous unapplied batches.

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
+pragma solidity ^0.5.0;
+
+import "./SnappBase.sol";
+
+
+contract SnappAuction is SnappBase {
```
fleupold, line 6: 😍

fleupold (commented):
Looks awesome.

My main concern is around the unconstrained growth of batches. Also would love to see PendingFlux renamed to something more general if reused for orders.

contracts/SnappAuction.sol
josojo: All these checks and require have the disadvantage that they cost gas for all users. While it is good for users, who make mistakes, its bad for all other users. I suggest to remove all these requires...

contracts/SnappAuction.sol
josojo: What do you mean by processing time takes longer? Isn't the processing time determined by the smart contract on how long it waits for "highest pids" for traders welfare? We just need to make sure that these two numbers, the 20 from here and the submission time for traders welfare are fitting together.

But for the escalation and stopping of chain, such a condition might make for sure a lot of sense

contracts/SnappAuction.sol
bh2smith: If someone submits an order with `tokenId` out of range, then the whole listener will break. For example, we will get IndexErrors on the balance array. Do we even know what the solver will do if the limit orders are between the same token pairs? Have we considered how to handle orders for non-existent token pairs?

I agree that this consumes gas, but where should this verification logic go? If it isn't handled at the source or error, then I presume we will have a lot of special cases to consider in the external services. 

I can remove all of this validation easily, but want to make sure we are all aware of the consequences we will have to face upon doing so.

contracts/SnappAuction.sol
bh2smith: I am in favour of `PendingBatch`

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    function encodeOrder(
+        uint16 accountId, 
+        uint8 buyToken, 
+        uint8 sellToken, 
+        uint128 buyAmount, 
+        uint128 sellAmount
+    ) internal pure returns (bytes32) {
+        // solhint-disable-next-line max-line-length
+        return bytes32(uint(accountId) + (uint(buyToken) << 16) + (uint(sellToken) << 24) + (uint(sellAmount) << 32) + (uint(buyAmount) << 132));
```
josojo, line 83: Should sellAmount and buyAmount be shifted by the same number?

contracts/SnappAuction.sol
bh2smith: What do you mean by "illegal" here? I was under the impression that "illegal" is defined by the failure to comply with the contract's requirements.

contracts/SnappAuction.sol
bh2smith: Good question. I wonder if this "pendingBatch" should have attributes for `prices` and `fulfilmentVolumes`. Then state transition can be performed based on the batch itself and the `appliedStateIndex` would actually make sense. This, however, would be a change to make in another PR (related to state transition from auction results).

test/snapp_auction.js
bh2smith: We already have 100% Branch Coverage. I made sure of this before submitting this PR.

contracts/SnappAuction.sol
bh2smith: I can get behind this. Will have to go back to the drawing board here though and think about a new approach.

contracts/SnappAuction.sol
josojo: You are right that it would be more annoying to deal with all these special cases in the listener, and also the driver would have to do the same. It's not so bad, but still unpleasant.

 Do we even know what the solver will do if the limit orders are between the same token pairs?
-> this can be added
Have we considered how to handle orders for non-existent token pairs?
-> user balances will be zero, so there can't happen any trade anyways

contracts/SnappAuction.sol
bh2smith: These numbers are shifted by the cumulative sum of the size of the previous number.

- buyToken by 16 (because accountID is uint16)
- sellToken by 24 = 16 + 8 (because buyToken is uint8)
- sellAmount by 32 = 24 + 8 (because sellToken is uint8)
- buyAmount by 132 = 32 + 100 (because sellAmount is uint100)

contracts/SnappAuction.sol
bh2smith: Good point. I will look into this and try to make it so.

contracts/SnappAuction.sol
josojo: Ben knows to shift :+1:

contracts/SnappAuction.sol
bh2smith: But if the tokenIds are greater than the max, we will get index errors when trying to access the balances....

fleupold: I guess our current choice of 30 tokens is not ideal. We should probably chose a power of 2 (e.g. 32) and then just trim all input so we can ensure the value is in range.

I'm also ok with leaving these requires for now.

contracts/SnappAuction.sol
fleupold: I wouldn't consider it illegal to trade GNO for GNO. I would consider it senseless but if someone wants to do it they are not attacking the contract in any malicious/illegal way.

If they found a way to manipulate the orders of other people, that I'd consider an example for illegality that should throw.

contracts/SnappAuction.sol
fleupold: If we see this data structure to be different from deposits/orders I'd suggest creating a separate struct and calling it PendingOrderBatch (and leaving PendingFlux). However if the structures will be the same for the foreseeable future, I think we should use the same type.

test/snapp_auction.js
fleupold: Sorry, I believe I meant 100% condition coverage. We are not checking the case where we creating a new batch because the old one is full.

contracts/SnappAuction.sol
fleupold: I'd suggest checking that for auctionIndex > 2, `auctions[auctionIndex -2].shaHash != 0`, which would imply that the auction has been applied.

contracts/SnappAuction.sol
fleupold: Although I'd still prefer if we could reuse code in solidity in better ways than inheritance.

test/snapp_auction.js
bh2smith: It will be difficult to get 1000 orders in one batch before 20 blocks are mined. Any suggestions on how would be very welcome.

contracts/SnappAuction.sol
bh2smith: I don't see how non-zero shaHash implies auction has been applied.

contracts/SnappAuction.sol
bh2smith: Is this solidity's problem, or something I should look into changing?

contracts/SnappAuction.sol
bh2smith: I will offload this discussion to the [Contract Optimization Wiki](https://github.com/gnosis/dex-contracts/wiki/Contract-Optimization-Opportunities)

contracts/SnappAuction.sol
bh2smith: Yes, this is true. It would make the contract and unit testing a lot simpler if some of these validations were lifted. However, it might not hurt to have all of this in place earlier, and then decide which can be removed. I suspect that the business logic of the external services would then have a lot more conditions to consider and handle. I suppose it all depends where these should be taken care of. I have included them here in the beginning, so to cut the problem off directly at the source. At least they are all recorded somewhere and their handling can be shuffled around at some point later...

contracts/SnappAuction.sol
bh2smith: At this moment (for this task), I see them as being the same. However, this could very-well change in the next task (when state transition resulting from auctions is implemented). Fortunately, that is not the purpose of this PR, so it will remain as is until a change is necessary.

It seems as if we may not want to store prices and volumes on the contract anyway, since this could, technically, be recovered from the transaction payload.

contracts/SnappAuction.sol
bh2smith: Why not simply reject all new orders until the previous batch has been applied? Or maybe allow for 3 batches to be collected before rejecting new orders.

I believe the condition for a batch having been applied is 

```
auctions[auctionIndex - 2].appliedAccountStateIndex != 0
```

(Now I see what you were referring to with the `auctions[auctionIndex - 2].shaHash` @fleupold ).

bh2smith: This should cover it

```js
if (auctionIndex > 1) {
    require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "To many pending auctions");
}
```

contracts/SnappAuction.sol
fleupold: solidity's problem

contracts/SnappAuction.sol
fleupold: Yes that's what I mean. Great!

test/snapp_auction.js
fleupold: That's because ganache by default mines a block on every transaction? You could create a smart contract just for the test that opens an account and adds 1000 orders in a single transaction. But not sure if that's worth the hassle.

On a side note, we are using amount of blocks passed rather than an actual measure of time. It's probably somewhat more reliable to use time directly (and create a new batch every 3 minutes). Block times tend to vary significantly and might at some point drastically increase due to the difficulty bomb.
Maybe something to do as part of the final cleanup in this milestone.

josojo: depending ontime rather than blocks also makes testing most of the time easier... sounds good to me.

contracts/SnappAuction.sol
josojo: I think it is good to allow GNO for GNO orders, as it is more gas efficient. But @twalth3r , your solver need to sort out these orders. Is your solver able to sort these order out or deal with them in any other meaning ful way?

test/snapp_auction.js
fleupold: Created #59

contracts/SnappAuction.sol
bh2smith: So, as I understand, we are waiting on this and then this PR might be ready for merge @fleupold @josojo ?

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,85 @@
...
+        uint auctionId, 
+        uint16 slotIndex,
+        uint16 accountId, 
+        uint8 buyToken, 
+        uint8 sellToken, 
+        uint128 buyAmount,
+        uint128 sellAmount
+    );
+
+    constructor () public {
```
josojo, line 23: Do you also have to call the constructor of the inherited contract?

contracts/SnappAuction.sol
fleupold: I don't understand why we need to wait for anything here? The solver is not going to be able to match an order where I sell 10 GNO for 11 GNO, because then the price would not be reflexive. It could match orders where we sell 10GNO for 10GNO and if someone wants to do that, I think he should be free to do so.

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    function placeSellOrder(
+        uint8 buyToken,
+        uint8 sellToken,
+        uint128 minBuyAmount,
+        uint128 maxSellAmount
+    ) public onlyRegistered() {
+        if (auctionIndex > 1) {
+            require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "To many pending auctions");
```
fleupold: I think this condition should go into the branch where we decide to create a new batch. Otherwise we are allowed to create the first order in the new batch but not more. Also I think having one batch fill up while the previous is not yet processed is fine, just not more than that.


Also is this condition covered by a unit test?

fleupold (changes requested):
One small request about the added require. Let's decide when we are trying to create a new batch if that is legal and add unit tests for this case.

josojo (approved):
Great! Good progress

When the inheritance constructor question is cleared and if we do not decide to make everything time depended instead of blocknr depended, then it's ready!

contracts/SnappAuction.sol
bh2smith: Nah, I checked this yesterday. Constructor of other contract only needs to be called if there are parameters to pass. We can trust in this because the unit tests haven't failed... although, we may want to update the migrations and change all the tests to check deposits on the `SnappAuction` This would really prove we don't need the constructor.

contracts/SnappAuction.sol
bh2smith: This condition will be covered by a unit test very soon.

Also, yes will move into new batch branch.

bh2smith:
@josojo The time and block dependence is now a separate task #59

contracts/SnappAuction.sol
josojo: makes sense

contracts/SnappAuction.sol
bh2smith: Done! Note that we can't get complete branch coverage until we have state transitions (i.e. a way of processing these batches). There is a unit test available for all that we can have at this stage though.

contracts/SnappAuction.sol
bh2smith: So is there a consensus to remove this constraint from the order validation? Someone just say so and its gone.

bh2smith: Forget it. Has been removed. Order are now allowed to have the same token pair.

contracts/SnappAuction.sol
```diff
@@ -0,0 +1,84 @@
...
+        require(tokenIdToAddressMap[sellToken] != address(0), "Sell token is not registered");
+
+        // Could also enforce that buyToken != sellToken, but not technically illegal.
+
+        if (
+            auctions[auctionIndex].size == AUCTION_BATCH_SIZE || 
+            block.number > auctions[auctionIndex].creationBlock + 20
+        ) {
+            if (auctionIndex > 1) {
+                require(auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions");
```
fleupold: ubernit: This can be one line
```
require(auctionIndex < 2 || auctions[auctionIndex - 2].appliedAccountStateIndex != 0, "Too many pending auctions");
```

fleupold (approved):
Very nice!

contracts/SnappAuction.sol
bh2smith: I prefer 1 liners.

bh2smith: But not gonna change it because it makes the line too wide.

fleupold: You could line-break

bh2smith: If you insist.

PR #82, opened 2019-04-25
Ensure SnappBase Emits Initialization event

bh2smith:


bh2smith:
This is not necessary.

PR #104, opened 2019-06-06
PR for discussion of implementing ordersubmission without nonce

josojo:


contracts/SnappAuction.sol
```diff
@@ -11,17 +11,11 @@ contract SnappAuction is SnappBase {
...
     }
 
-    struct StandingOrderData {
-        mapping(uint => StandingOrderBatch) reservedAccountOrders;
-        uint standingOrderNonce;
-    }
-
-    //mapping from accountId to nonce to StandingOrderBatch
-    mapping (uint16 => StandingOrderData) public standingOrderPlacements;
+    //mapping from accountId to validFromAuctionIndex to StandingOrderBatch
```
bh2smith: can't this just be standing orders? I don't see the reason for the word "Placement" here

contracts/SnappAuction.sol
```diff
@@ -11,17 +11,11 @@ contract SnappAuction is SnappBase {
...
+    mapping (uint16 => mapping(uint => StandingOrderBatch)) public standingOrderPlacements;
```
bh2smith, line 18: I don't see the reason for the word "Placements" here. Can we please just have "standingOrders"

bh2smith (commented):
I really don't like all these long words: Here are my suggestions in order of priority.

- prevValidFromAuctionIndex --> prevAuctionFrom
- standingOrderPlacements --> standingOrders
- validFromAuctionIndex --> auctionFrom
- validToAuctionIndex --> auctionTo

bh2smith:
Also, if we are going to start using `userId` I would think that every use of `accountId` should also be changed everywhere. That is, we need to use one or the other consistently, but not a mix of both. No strong preference either way.

contracts/SnappAuction.sol
```diff
@@ -79,21 +73,14 @@ contract SnappAuction is SnappBase {
         return auctions[slot].appliedAccountStateIndex != 0;
     }
 
-    function getStandingOrderHash(uint16 userId, uint128 nonce) public view returns (bytes32) {
-        return standingOrderPlacements[userId].reservedAccountOrders[nonce].orderHash;
+    function getStandingOrderHash(uint16 userId, uint validFromAuctionIndex) public view returns (bytes32) {
```
bh2smith, line 76: I don't see that this function is actually invoked anywhere. Is it needed?

contracts/SnappAuction.sol
```diff
@@ -79,21 +73,14 @@ contract SnappAuction is SnappBase {
...
+        return standingOrderPlacements[userId][validFromAuctionIndex].orderHash;
     }
     
-    function getStandingOrderValidFrom(uint16 userId, uint128 nonce) public view returns (uint) {
-        return standingOrderPlacements[userId].reservedAccountOrders[nonce].validFromAuctionIndex;
+    function getStandingOrdervalidToAuctionIndex(uint16 userId, uint validFromAuctionIndex) public view returns (uint) {
```
bh2smith, line 80: Same. Is this required for anything? Do we plan to use it for unit testing?

contracts/SnappAuction.sol
```diff
@@ -102,13 +89,17 @@ contract SnappAuction is SnappBase {
...
+        uint128[] memory sellAmounts,
+        uint prevValidFromAuctionIndex
     ) public onlyRegistered() {
         
         // Update Auction Hash based on request
         uint16 accountId = publicKeyToAccountMap(msg.sender);
         require(accountId <= AUCTION_RESERVED_ACCOUNTS, "Accout is not a reserved account");
 
+        // Check correctness of prevValidFromAuctionIndex
+        require( standingOrderPlacements[accountId][prevValidFromAuctionIndex].validToAuctionIndex != MAX_UINT || prevValidFromAuctionIndex == 0, "validToAuctionIndex is set, this is not the previous orderbatch");
```
bh2smith, line 102:
```suggestion
        require(standingOrderPlacements[accountId][prevValidFromAuctionIndex].validToAuctionIndex != MAX_UINT || prevValidFromAuctionIndex == 0, "validToAuctionIndex is set, this is not the previous orderbatch");
```

contracts/SnappAuction.sol
```diff
@@ -127,17 +118,13 @@ contract SnappAuction is SnappBase {
...
-            standingOrderPlacements[accountId].reservedAccountOrders[currentNonce].validToAuctionIndex = auctionIndex - 1;
-            newNonce += 1;
-            standingOrderPlacements[accountId].standingOrderNonce = newNonce;
-            standingOrderPlacements[accountId].reservedAccountOrders[newNonce].validFromAuctionIndex = auctionIndex;
+        if (auctionIndex > prevValidFromAuctionIndex) {
+            standingOrderPlacements[accountId][prevValidFromAuctionIndex].validToAuctionIndex = auctionIndex - 1;
+            standingOrderPlacements[accountId][auctionIndex].validToAuctionIndex = MAX_UINT;
         }
-        standingOrderPlacements[accountId].reservedAccountOrders[newNonce].orderHash = orderHash;
+        standingOrderPlacements[accountId][auctionIndex].orderHash = orderHash;
```
bh2smith, line 132: This looks like a weird/inconsistent change that does not reflect identical logic to the previous code (which used an apparently different value).

josojo: In my head this makes sense :) 
the mapping was changed to 
accoutID => validFromAuctionIndex => StandingOrderBatch

And the new validity will be set here to (validityFromAuctionIndex= auctionId)

This all goes a little bit further than felix suggestion, as it basically removes the nonce and stores the validFrom only indirectly.

bh2smith: Cool, just making sure!

Issue #96, opened 2019-06-04
Unit test completely filled batches


bh2smith:
Could be achieved by first allowing multiple orders to be submitted in a single transaction.

bh2smith:
closed by #120

PR #118, opened 2019-06-11
#96 - Unit Testing Full Auction Batches

bh2smith:
This test demonstrates that the intended functionality happens for the "regular orders" in a filled auction batch (of size 500). While testing this I realized that special orders can't close auction batches since these orders continually overwrite the previous, while not occupying any additional space in the batch. This means that the only way an auction batch can close early is when the regular order space if full.

### DIFF
--- a/.solhintignore
+++ b/.solhintignore
@@ -1,5 +1,7 @@
 contracts/Migrations.sol
 .solhint.json
 .solhintignore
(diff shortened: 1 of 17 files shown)


bh2smith:
Closing this and will reopen with cleaner commit history.

PR #122, opened 2019-06-12
Redundant Requirement

bh2smith:
Now that sell and buy Amounts are `uint96` by default no need to check that they are passed with at most 96 bits.


fleupold (approved):
🦂

PR #128, opened 2019-06-12
Emits event standingSellOrder with packedOrder data 

josojo:
Emits event standingSellOrder with packedOrder data as originally requested by Felix

testplan:
unit tests

### DIFF
--- a/test/snapp_auction.js
+++ b/test/snapp_auction.js
@@ -538,7 +538,7 @@ contract("SnappAuction", async (accounts) => {
       assert.equal(currentAuction.size, 2)
     })
 
-    it.only("Encodes order information in emitted event", async () => {
+    it("Encodes order information in emitted event", async () => {
       const instance = await SnappAuction.new()
       await setupEnvironment(MintableERC20, instance, token_owner, [user_1], 2)
       const order = encodeOrder(0, 1, 2, 3)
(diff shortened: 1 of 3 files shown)


fleupold (approved):
👏 

Thanks, this should unblock the listener PR.

test/snapp_auction.js
```diff
@@ -524,12 +544,10 @@ contract("SnappAuction", async (accounts) => {
       const order = encodeOrder(0, 1, 2, 3)
       const tx = await instance.placeSellOrders(order, { from: user_1 })
       const eventLog = tx.logs
-
-      const buyToken = eventLog[0].args.buyToken
       const sellToken = eventLog[0].args.sellToken
+      const buyToken = eventLog[0].args.buyToken
```
bh2smith: I don't see why this was switched! Would prefer if it was switched back.

test/snapp_auction.js
```diff
@@ -10,7 +10,7 @@ const {
 
 const {
   isActive,
-  encodeOrder }  = require("./snapp_utils.js")
+  encodeOrder}  = require("./snapp_utils.js")
```
bh2smith:
```suggestion
  encodeOrder }  = require("./snapp_utils.js")
```

bh2smith (approved):
Looks good, I only had a few tiny comments about some unnecessary changes that don't really contribute to the point of the PR. Very cool!

PR #142, opened 2019-06-26
eth-gas-reporter debugging...

cgewecke:
This isn't a real PR. Can't reproduce [eth-gas-reporter 139](https://github.com/cgewecke/eth-gas-reporter/issues/139) locally on OSX or in Travis CI over there so will try a few things here...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - '8'
+  - '10'
 before_install:
   - rm -rf build
   - rm -rf node_modules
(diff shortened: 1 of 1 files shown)


cgewecke:
Looks like using yarn is one workaround. Will look into this further though. . .

See [Travis Build 792](https://travis-ci.org/gnosis/dex-contracts/builds/550711078?utm_source=github_status&utm_medium=notification)

bh2smith:
Thanks @cgewecke for looking into this. I saw you mentioned "try shrinkwrap" but I don't think I know what that means. The workaround seems like it should be alright for now.

cgewecke:
@bh2smith Ah "check shrinkwrap" was just a note-to-self about trying understand why npm is having trouble with the BigNumber git fork. Glad that worked out!

PR #143, opened 2019-06-27
Update Bytes Utils

bh2smith:
Recently Bytes-Utils has included functionality for converting byte slices to `uint96` so that we no longer need to put assembly in our code.


josojo (approved):
Looks good. 
Great that you improved the library!

Just the npm install issue needs to be fixed

PR #148, opened 2019-07-09
Introduce coveralls to generate coverage badge

bh2smith:


fleupold (approved):
Cool, is this something we can add to the services repo as well?

bh2smith:
> Cool, is this something we can add to the services repo as well?

Yes I would certainly hope to do so, but I will have to investigate how to do such a thing in RUST/Python project...

PR #164, opened 2019-08-09
First step solution submission: adapting orders and balances

josojo:
This pr allows a solution provider to submit their solution;

The function processes the solution, under the assumption that it was provided according to the time constraints and updates all balances and orders touched in a solution.

BuyOrders can not yet be submitted.

Tom already mentioned that we might wanna replace some equalities with delta checks, in order to make solution submission possible for him. But this will be postponed.

Testplan:
unit test

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -110,12 +110,13 @@ contract StablecoinConverter is EpochTokenLocker {
         return IdToAddressBiMap.getAddressAt(registeredTokens, id);
     }
 
+    mapping (uint16 => uint128) public currentPrices;
     struct TradeData {
         address owner;
         uint volume;
         uint16 orderIds;
     }
-    mapping (uint16 => uint128) public currentPrice;
+
     function submitSolution(
         uint32 batchIndex,
         address[] memory owners,  //tradeData is submitted as arrays
@@ -136,10 +137,10 @@ contract StablecoinConverter is EpochTokenLocker {
             require(order.validUntil >= batchIndex, "Order is no longer valid");
             // Assume for now that we always have sellOrders
             uint128 executedSellAmount = volumes[i];
-            require(currentPrice[order.sellToken] != 0, "prices are not allowed to be zero");
+            require(currentPrices[order.sellToken] != 0, "prices are not allowed to be zero");
             uint128 executedBuyAmount = uint128(
-                volumes[i].mul(currentPrice[order.buyToken]) /
-                currentPrice[order.sellToken]
+                volumes[i].mul(currentPrices[order.buyToken]) /
+                currentPrices[order.sellToken]
             );
             // Ensure executed price is not lower than the order price:
             //       executedSellAmount / executedBuyAmount >= order.sellAmount / order.buyAmount
@@ -168,12 +169,12 @@ contract StablecoinConverter is EpochTokenLocker {
         orders[owner][orderId].sellAmount = newSellAmount;
     }
 
-    function writeCurrentPrices(
+    function CurrentPrices(
         uint128[] memory prices,  //list of prices for touched token only
         uint16[] memory tokenIdsForPrice  // price[i] is the price for the token with tokenID tokenIdsForPrice[i]
     ) internal {
         for (uint i = 0; i < tokenIdsForPrice.length; i++) {
-            currentPrice[tokenIdsForPrice[i]] = prices[i];
+            currentPrices[tokenIdsForPrice[i]] = prices[i];
         }
     }
 }
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -115,7 +115,7 @@ contract StablecoinConverter is EpochTokenLocker {
         uint volume;
         uint16 orderIds;
     }
-
+    mapping (uint16 => uint128) public currentPrice;
     function submitSolution(
         uint32 batchIndex,
         address[] memory owners,  //tradeData is submitted as arrays
@@ -128,6 +128,7 @@ contract StablecoinConverter is EpochTokenLocker {
             batchIndex == getCurrentStateIndex() - 1,
             "Solutions are no longer accepted for this batch"
         );
+        writeCurrentPrices(prices, tokenIdsForPrice);
         uint len = owners.length;
         for (uint i = 0; i < len; i++) {
             Order memory order = orders[owners[i]][orderIds[i]];
@@ -135,21 +136,19 @@ contract StablecoinConverter is EpochTokenLocker {
             require(order.validUntil >= batchIndex, "Order is no longer valid");
             // Assume for now that we always have sellOrders
             uint128 executedSellAmount = volumes[i];
+            require(currentPrice[order.sellToken] != 0, "prices are not allowed to be zero");
             uint128 executedBuyAmount = uint128(
-                volumes[i].mul(prices[findPriceIndex(order.buyToken, tokenIdsForPrice)]) /
-                prices[findPriceIndex(order.sellToken, tokenIdsForPrice)]
+                volumes[i].mul(currentPrice[order.buyToken]) /
+                currentPrice[order.sellToken]
             );
+            // Ensure executed price is not lower than the order price:
+            //       executedSellAmount / executedBuyAmount >= order.sellAmount / order.buyAmount
             require(
                 executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
                 "limit price not satisfied"
             );
             require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
-            uint128 newSellAmount = uint128(order.sellAmount.sub(executedSellAmount));
-            orders[owners[i]][orderIds[i]].buyAmount = uint128(
-                newSellAmount
-                .mul(order.buyAmount) / order.sellAmount
-            );
-            orders[owners[i]][orderIds[i]].sellAmount = newSellAmount;
+            updateRemainingOrder(owners[i], orderIds[i], executedSellAmount);
             addBalance(owners[i], tokenIdToAddressMap(order.buyToken), executedBuyAmount);
         }
         // doing all subtractions after all additions (in order to avoid negative values)
@@ -159,13 +158,22 @@ contract StablecoinConverter is EpochTokenLocker {
         }
     }
 
-    function findPriceIndex(uint16 index, uint16[] memory tokenIdsForPrice) public pure returns (uint) {
-        uint length = tokenIdsForPrice.length;
-        for (uint i = 0; i < length; i++) {
-            if (tokenIdsForPrice[i] == index) {
-                return i;
-            }
+    function updateRemainingOrder(address owner, uint orderId, uint128 executedSellAmount) internal returns (uint) {
+        Order memory order = orders[owner][orderId];
+        uint128 newSellAmount = uint128(order.sellAmount.sub(executedSellAmount));
+        orders[owner][orderId].buyAmount = uint128(
+            newSellAmount
+            .mul(order.buyAmount) / order.sellAmount
+        );
+        orders[owner][orderId].sellAmount = newSellAmount;
+    }
+
+    function writeCurrentPrices(
+        uint128[] memory prices,  //list of prices for touched token only
+        uint16[] memory tokenIdsForPrice  // price[i] is the price for the token with tokenID tokenIdsForPrice[i]
+    ) internal {
+        for (uint i = 0; i < tokenIdsForPrice.length; i++) {
+            currentPrice[tokenIdsForPrice[i]] = prices[i];
         }
-        revert("Price not provided for token");
     }
 }
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -121,8 +121,8 @@ contract StablecoinConverter is EpochTokenLocker {
         address[] memory owner,  //tradeData is submitted as arrays
         uint16[] memory orderId,
         uint128[] memory volume,
-        uint128[] memory prices,
-        uint16[] memory tokenIdForPrice
+        uint128[] memory prices,  //list of prices for touched token only
+        uint16[] memory tokenIdForPrice  // price[i] is the price for the token with tokenID tokenIdForPrice[i]
     ) public {
         require(
             batchIndex == getCurrentStateIndex() - 1,
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -144,9 +144,9 @@ contract StablecoinConverter is EpochTokenLocker {
                 currentPrices[order.sellToken]
             );
             // Ensure executed price is not lower than the order price:
-            //       executedSellAmount / executedBuyAmount >= order.sellAmount / order.buyAmount
+            //       executedSellAmount / executedBuyAmount <= order.sellAmount / order.buyAmount
             require(
-                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                executedSellAmount.mul(order.buyAmount) <= executedBuyAmount.mul(order.sellAmount),
                 "limit price not satisfied"
             );
             require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -155,8 +155,11 @@ contract StablecoinConverter is EpochTokenLocker {
         }
         // doing all subtractions after all additions (in order to avoid negative values)
         for (uint i = 0; i < len; i++) {
-            Order memory order = orders[owners[i]][orderIds[i]];
-            subtractBalance(owners[i], tokenIdToAddressMap(order.sellToken), volumes[i]);
+            subtractBalance(
+                owners[i],
+                tokenIdToAddressMap(orders[owners[i]][orderIds[i]].sellToken),
+                volumes[i]
+            );
         }
     }
 
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -128,36 +128,33 @@ contract StablecoinConverter is EpochTokenLocker {
             batchIndex == getCurrentStateIndex() - 1,
             "Solutions are no longer accepted for this batch"
         );
-        // checkTradersUtility();
-        // checkPriceConformity(prices);
-        // checkNorm(prices);
         uint len = owner.length;
         for (uint i = 0; i < len; i++) {
             Order memory order = orders[owner[i]][orderId[i]];
             require(order.validFrom <= batchIndex, "Order is not yet valid");
             require(order.validUntil >= batchIndex, "Order is no longer valid");
             // Assume for now that we always have sellOrders
-            uint128 sellVolume = volume[i];
-            uint128 buyVolume = uint128(
+            uint128 executedSellAmount = volume[i];
+            uint128 executedBuyAmount = uint128(
                 volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
                 prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
             );
             require(
-                sellVolume.mul(order.buyAmount) >= buyVolume.mul(order.sellAmount),
-                "limit price not met"
+                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                "limit price not satisfied"
             );
-            require(order.sellAmount >= sellVolume, "sellVolume bigger than specified in order");
+            require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
             orders[owner[i]][orderId[i]].buyAmount = uint128(
-                order.sellAmount.sub(sellVolume)
+                order.sellAmount.sub(executedSellAmount)
                 .mul(order.buyAmount) / order.sellAmount
             );
-            orders[owner[i]][orderId[i]].sellAmount = uint128(order.sellAmount.sub(sellVolume));
-            addBalance(owner[i], tokenIdToAddressMap(order.buyToken), buyVolume);
+            orders[owner[i]][orderId[i]].sellAmount = uint128(order.sellAmount.sub(executedSellAmount));
+            addBalance(owner[i], tokenIdToAddressMap(order.buyToken), executedBuyAmount);
         }
-        //doing the substracts after doing all additions, in order to avoid negative values
+        //doing the subtracts after doing all additions, in order to avoid negative values
         for (uint i = 0; i < len; i++) {
             Order memory order = orders[owner[i]][orderId[i]];
-            substractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
+            subtractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
         }
     }
 
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -154,7 +154,7 @@ contract StablecoinConverter is EpochTokenLocker {
             orders[owner[i]][orderId[i]].sellAmount = uint128(order.sellAmount.sub(sellVolume));
             addBalance(owner[i], tokenIdToAddressMap(order.buyToken), buyVolume);
         }
-        //doing the substracts after doing all additions, in order to avoid negative values
+        // doing all subtractions after all additions (in order to avoid negative values)
         for (uint i = 0; i < len; i++) {
             Order memory order = orders[owner[i]][orderId[i]];
             substractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -111,6 +111,7 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     mapping (uint16 => uint128) public currentPrices;
+
     struct TradeData {
         address owner;
         uint volume;
@@ -169,7 +170,7 @@ contract StablecoinConverter is EpochTokenLocker {
         orders[owner][orderId].sellAmount = newSellAmount;
     }
 
-    function CurrentPrices(
+    function writeCurrentPrices(
         uint128[] memory prices,  //list of prices for touched token only
         uint16[] memory tokenIdsForPrice  // price[i] is the price for the token with tokenID tokenIdsForPrice[i]
     ) internal {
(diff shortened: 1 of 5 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -17,8 +17,8 @@ contract StablecoinConverter is EpochTokenLocker {
         bool isSellOrder,
         uint32 validFrom,
         uint32 validUntil,
-        uint256 buyAmount,
-        uint256 sellAmount
+        uint128 buyAmount,
```
bh2smith, line 20: Are you sure we don't want `uint96` here (like we do in the other contracts)?

contracts/StablecoinConverter.sol
```diff
@@ -41,6 +41,7 @@ contract StablecoinConverter is EpochTokenLocker {
     mapping(address => Order[]) public orders;
 
     IdToAddressBiMap.Data private registeredTokens;
+
     uint public MAX_TOKENS; // solhint-disable-line
```
bh2smith, line 45: Would prefer to know what this is disabling...

I believe you can write the specific item right after.

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        address owner;
+        uint volume;
+        uint16 orderId;
+    }
+
+    function submitSolution(
+        uint32 batchIndex,
+        address[] memory owner,  //tradeData is submitted as arrays
+        uint16[] memory orderId,
+        uint128[] memory volume,
```
bh2smith: Don't we need both `buyAmount` and `sellAmount`?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint128[] memory prices,
+        uint16[] memory tokenIdForPrice
+    ) public {
+        require(
+            batchIndex == getCurrentStateIndex() - 1,
+            "Solutions are no longer accepted for this batch"
+        );
+        // checkTradersUtility();
+        // checkPriceConformity(prices);
+        // checkNorm(prices);
```
bh2smith: Are these TODOs?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
+            uint128 sellVolume = volume[i];
+            uint128 buyVolume = uint128(
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
+            );
+            require(
+                sellVolume.mul(order.buyAmount) >= buyVolume.mul(order.sellAmount),
+                "limit price not met"
```
```suggestion
                "limit price not satisfied"
```

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        // checkPriceConformity(prices);
+        // checkNorm(prices);
+        uint len = owner.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            require(order.validFrom <= batchIndex, "Order is not yet valid");
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
+            uint128 sellVolume = volume[i];
+            uint128 buyVolume = uint128(
```
bh2smith: Are we being consistent with the fact that `Volume = Price * Amount`.

Is this "buyVolume" supposed to represent the "executedBuyAmount" provided in the solution?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+                order.sellAmount.sub(sellVolume)
+                .mul(order.buyAmount) / order.sellAmount
+            );
+            orders[owner[i]][orderId[i]].sellAmount = uint128(order.sellAmount.sub(sellVolume));
+            addBalance(owner[i], tokenIdToAddressMap(order.buyToken), buyVolume);
+        }
+        //doing the substracts after doing all additions, in order to avoid negative values
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            substractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
```
bh2smith: nit "subtract"!

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            );
+            require(order.sellAmount >= sellVolume, "sellVolume bigger than specified in order");
+            orders[owner[i]][orderId[i]].buyAmount = uint128(
+                order.sellAmount.sub(sellVolume)
+                .mul(order.buyAmount) / order.sellAmount
+            );
+            orders[owner[i]][orderId[i]].sellAmount = uint128(order.sellAmount.sub(sellVolume));
+            addBalance(owner[i], tokenIdToAddressMap(order.buyToken), buyVolume);
+        }
+        //doing the substracts after doing all additions, in order to avoid negative values
```
```suggestion
        // doing all subtractions after all additions (in order to avoid negative values)
```

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,65 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            substractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
+        }
+    }
+
+    function findPriceIndex(uint16 index, uint16[] memory tokenIdForPrice) public pure returns (uint) {
```
bh2smith: This is cute, but somewhat inefficient...

Observe that you are fetching the price index for each order in the loop above (`submitSolution`). This means you have `O(numTokens * numOrders)` complexity while you could cache this token index info once and continue reusing it! This would reduce your `submitSolution` to `O(numOrders)`

bh2smith (dismissed):
Looks pretty cool, I would suggest reworking the usage of `findPriceIndex` as stated in a comment above!

contracts/StablecoinConverter.sol
josojo: yes, with uint128, we can store the order placement with 2 word (512 bit)stores in the EVM. reducing to uint96 would not change the picture.

contracts/StablecoinConverter.sol
josojo: Hopefully not. Tom will look into it more in-depth.

contracts/StablecoinConverter.sol
josojo: I agree that it is not highly inefficient. 

But I would like to continue like that for now and make this optimization later. The reason is that later on, once the function submitSolution is feature complete, I expect to have trouble with the stack deepness. I can only have 16 variables and I am already at 10.  Checking which values should be stored and which one should be recomputed should be done, once we have everything well defined.

bh2smith: Ok then. Maybe we could include a comment or todo so this doesn’t get buried.

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+
+    struct TradeData {
+        address owner;
+        uint volume;
+        uint16 orderId;
+    }
+
+    function submitSolution(
+        uint32 batchIndex,
+        address[] memory owner,  //tradeData is submitted as arrays
```
fleupold: I think it would be nice to pass structs of TradeData here. Managing separate arrays which elements have to align is much more messy than having one array of coherent items. Is this possible?

This would also reduce the numbers of variables you are working with.

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint16[] memory orderId,
+        uint128[] memory volume,
+        uint128[] memory prices,
```
fleupold: Can you explain this data structure in combination with the one below a bit more? Could this be a list where the index signifies the token? Or a mapping or tokenId to price?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint len = owner.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            require(order.validFrom <= batchIndex, "Order is not yet valid");
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
+            uint128 executedSellAmount = volume[i];
+            uint128 executedBuyAmount = uint128(
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
```
fleupold: How awesome would it be if this could just be `prices[order.sellToken]`

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint128[] memory prices,
+        uint16[] memory tokenIdForPrice
+    ) public {
+        require(
+            batchIndex == getCurrentStateIndex() - 1,
+            "Solutions are no longer accepted for this batch"
+        );
+        uint len = owner.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
```
fleupold: Should this be storage so that it can be used below instead of always referring to `orders[owner[i]][orderId[i]]` which is very hard to read.

test/stablecoin_converter.js
```diff
@@ -107,6 +111,390 @@ contract("StablecoinConverter", async (accounts) => {
...
+      
+      const tokenIdForPrice = [0, 1]
+
+      //correct batchIndex would be batchIndex
+      await truffleAssert.reverts(
+        stablecoinConverter.submitSolution(batchIndex - 1, owner, orderId, volume,  prices, tokenIdForPrice),
+        "Order is not yet valid"
+      )
+    })
+    it("throws, if order is not yet valid", async () => {
```
fleupold: "throws, if order is no longer valid"

fleupold (commented):
Nice unit tests. Some comments inline.

contracts/StablecoinConverter.sol
josojo: Completely agree. But this is only available with the experimental abiEncoder and I do not want to rely on it

contracts/StablecoinConverter.sol
josojo: But storage reads are more expensive than storing it once in memory and reading it from there, isn't it?

contracts/StablecoinConverter.sol
josojo: > list where the index signifies the token?

Well, if we allow anyone to add tokens, then this is too much. Then we would have an array of size 2*16  in the worst case 

> mapping
Yeah, something like this would be great. But mappings  can not be memory data types and can not be provided as function parameters

The easiest would probably to provide a price per order, but then checking price arbitrage freeness would become a hassle.

 I don't see a way to hand the data over to the smart contract in a smarter way. 
Just as Ben mentioned, we could store it in memory in the smart contract easier...

contracts/StablecoinConverter.sol
anxolin: nit: Should be MAX_TOKENS a constant instead of being initialized in the constructor?
It's cheaper and we don't allow changing the tokens anyways. Or is maybe for testing?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        address[] memory owner,  //tradeData is submitted as arrays
+        uint16[] memory orderId,
+        uint128[] memory volume,
+        uint128[] memory prices,  //list of prices for touched token only
+        uint16[] memory tokenIdForPrice  // price[i] is the price for the token with tokenID tokenIdForPrice[i]
+    ) public {
+        require(
+            batchIndex == getCurrentStateIndex() - 1,
+            "Solutions are no longer accepted for this batch"
+        );
```
anxolin, line 132: Shouldn't we do some validations on the sizes of the arrays? Or the idea is to save some gas?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            uint128 executedBuyAmount = uint128(
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
+            );
+            require(
+                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                "limit price not satisfied"
+            );
+            require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
+            orders[owner[i]][orderId[i]].buyAmount = uint128(
```
anxolin: Can owner[i], orderId[i] be refactored as you did with:

`uint128 executedSellAmount = volume[i];`

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+    struct TradeData {
+        address owner;
+        uint volume;
+        uint16 orderId;
+    }
+
+    function submitSolution(
+        uint32 batchIndex,
+        address[] memory owner,  //tradeData is submitted as arrays
+        uint16[] memory orderId,
```
anxolin: All this params are arrays, but their name is in singular

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            "Solutions are no longer accepted for this batch"
+        );
+        uint len = owner.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            require(order.validFrom <= batchIndex, "Order is not yet valid");
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
+            uint128 executedSellAmount = volume[i];
+            uint128 executedBuyAmount = uint128(
```
anxolin, line 142: nit: Maybe small comments with the formula would make it easier to read when there's formulas: 
`buyAmount = volume x priceBuy / priceSell`

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
+            );
+            require(
```
anxolin, line 148: This is correct, but is not more readable like this?
`executedSellAmount / executedBuyAmount >= order.sellAmount / order.buyAmount`

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        // doing all subtractions after all additions (in order to avoid negative values)
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owner[i]][orderId[i]];
+            subtractBalance(owner[i], tokenIdToAddressMap(order.sellToken), volume[i]);
+        }
+    }
+
+    function findPriceIndex(uint16 index, uint16[] memory tokenIdForPrice) public pure returns (uint) {
+        uint length = tokenIdForPrice.length;
+        for (uint i = 0; i < length; i++) {
```
anxolin: The linter complains about `i` and `order` that where previously declared:

image

contracts/StablecoinConverter.sol
anxolin: Also, I think `orders[owner[i]][orderId[i]]` is can be substituted by `order`, it was already saved in a local var

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
+            );
+            require(
+                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                "limit price not satisfied"
+            );
+            require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
+            orders[owner[i]][orderId[i]].buyAmount = uint128(
+                order.sellAmount.sub(executedSellAmount)
```
anxolin: Is it worth it to save this in a local var? 
`order.sellAmount.sub(executedSellAmount)`

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,62 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            uint128 executedSellAmount = volume[i];
+            uint128 executedBuyAmount = uint128(
+                volume[i].mul(prices[findPriceIndex(order.buyToken, tokenIdForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdForPrice)]
+            );
+            require(
+                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                "limit price not satisfied"
+            );
+            require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
```
anxolin, line 152: Where do we plan to make sure the user has enough balance?
The orders can be arbitrarily big, I understand the solver needs to be aware also about the balance of the users for the slot, so it uses the `Min(balance, sellVolume)` in the optimization problem.

So I understand we would need to check in the submission of solutions...

anxolin (commented):
Very good one! I only added some comments, mostly nits.

The part where we score how good is the solution will be in a separate PR, right?

contracts/StablecoinConverter.sol
josojo: Yes, it is needed for testing. If we want to shoot 2**16 addToken requests for testing the limit, it takes too much time.

contracts/StablecoinConverter.sol
josojo: I think the validation is simply not needed. It does not provide any safety.

People can provide a false length array [a1, a2, ..., ak], even if we check for the size, by simply handing over [a1, a2,...ak, 0,...,0].
It would be security that could easily be circumvented. I hope that the limit price checks and token conservation checks will be sufficient to determine good solutions.

contracts/StablecoinConverter.sol
josojo: `orders[owner[i]][orderId[i]]` was saved as memory, but here, we want to update the stored value in the smart contract.

josojo: > owner[i], orderId[i] be refactored

agreed. I would like to refactor all variable names, once the solutionSubmission function is ready. The issue is that I have only 16 variables (because of stack deepness) and I am not yet sure, how to use them the best.

contracts/StablecoinConverter.sol
josojo: It is already happening in the substractBalance function from the epochTokenLocker. In there we check for underflows.

contracts/StablecoinConverter.sol
josojo: Yeah, but this formulation is prone to rounding errors. I prefer the multiplication.

contracts/StablecoinConverter.sol
josojo: Hm... not sure why your linter is complaining like this. Why should it be a bad pattern? Should I call the variables differently?

contracts/StablecoinConverter.sol
josojo: These are all good comments. I want to do this kind of refactoring later, how do I preserve them the best...?

josojo: yeah, I did it. It makes perfect sense and can be reverted later, in case we have not enough variables.

contracts/StablecoinConverter.sol
anxolin: I understand. Makes sense

contracts/StablecoinConverter.sol
anxolin: Ok, I understand.

It's not usually a big deal, just this small things (comments & refactors) can helps reviewers to spot errors, but if you think you'll be moving them back and forth when you add new things is understandable.

contracts/StablecoinConverter.sol
anxolin: Yep, that's right.

What about a small comment then?

i.e.
```
// Ensure executed price is not worse than the order price:
//       executedSellAmount / executedBuyAmount >= order.sellAmount / order.buyAmount
//       executedSellAmount * order.buyAmount >= order.sellAmount * executedBuyAmount
```

contracts/StablecoinConverter.sol
anxolin: Not sure, I think in the past (`v0.4`) was not allowed and this is why my current linter complains.

It seams that's ok after `0.5`

contracts/StablecoinConverter.sol
fleupold: Incoherent lengths wouldn't be an issue if we passed a single array with packed bytes for each entry (20 bytes for address, 2 bytes for ownerId, 16 bytes for volume, etc), which could either be parsed into an array of TradeData struct or accessed from via helper methods (similar to [here](https://github.com/gnosis/dex-contracts/blob/on_chain_verifier/contracts/SnappAuctionChallenge.sol#L270-L273)). 

This would also save you variables.

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,63 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint len = owners.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owners[i]][orderIds[i]];
+            require(order.validFrom <= batchIndex, "Order is not yet valid");
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
+            uint128 executedSellAmount = volumes[i];
+            uint128 executedBuyAmount = uint128(
+                volumes[i].mul(prices[findPriceIndex(order.buyToken, tokenIdsForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdsForPrice)]
```
fleupold: Given you will be writing the price list in SC storage anyways (in the next PR), would it make sense to store it as a mapping on storage and then access it in O(1) instead of doing a linear search over it?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,63 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        }
+        // doing all subtractions after all additions (in order to avoid negative values)
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owners[i]][orderIds[i]];
+            subtractBalance(owners[i], tokenIdToAddressMap(order.sellToken), volumes[i]);
+        }
+    }
+
+    function findPriceIndex(uint16 index, uint16[] memory tokenIdsForPrice) public pure returns (uint) {
+        uint length = tokenIdsForPrice.length;
```
fleupold: If we decide to use this for the long term, we should probably binary search in here and require that the list is passed sorted by tokenId.

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,63 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            orders[owners[i]][orderIds[i]].buyAmount = uint128(
+                newSellAmount
+                .mul(order.buyAmount) / order.sellAmount
+            );
+            orders[owners[i]][orderIds[i]].sellAmount = newSellAmount;
+            addBalance(owners[i], tokenIdToAddressMap(order.buyToken), executedBuyAmount);
+        }
+        // doing all subtractions after all additions (in order to avoid negative values)
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owners[i]][orderIds[i]];
```
fleupold: Not sure how this works in detail, but you are loading the entire order (2 words) into memory but are only reading 1 word from it. Maybe reading directly from storage is cheaper in that case?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,63 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        require(
+            batchIndex == getCurrentStateIndex() - 1,
+            "Solutions are no longer accepted for this batch"
+        );
+        uint len = owners.length;
+        for (uint i = 0; i < len; i++) {
+            Order memory order = orders[owners[i]][orderIds[i]];
+            require(order.validFrom <= batchIndex, "Order is not yet valid");
+            require(order.validUntil >= batchIndex, "Order is no longer valid");
+            // Assume for now that we always have sellOrders
```
fleupold, line 139: Should we assert that assumption in the place order method?

contracts/StablecoinConverter.sol
```diff
@@ -108,4 +109,63 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            uint128 executedBuyAmount = uint128(
+                volumes[i].mul(prices[findPriceIndex(order.buyToken, tokenIdsForPrice)]) /
+                prices[findPriceIndex(order.sellToken, tokenIdsForPrice)]
+            );
+            require(
+                executedSellAmount.mul(order.buyAmount) >= executedBuyAmount.mul(order.sellAmount),
+                "limit price not satisfied"
+            );
+            require(order.sellAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
+            uint128 newSellAmount = uint128(order.sellAmount.sub(executedSellAmount));
```
fleupold: nit: Maybe extract this into a private function `updateRemainingOrder` to document and make it a bit easier to understand the already pretty large `submitSolution` method?

fleupold (approved):
As mentioned in a PR a few weeks ago, I'd prefer a single array with packed data over multiple arrays which are implicitly expected to align. But up to you. Otherwise I think this looks good.

bh2smith (commented):
Getting better every day!

josojo:
I think this PR is ready for merging. The refactoring with blobs instead of array lists will be done in a separate PR. @bh2smith can you take another look, as your approval is blocking the merge?

PR #167, opened 2019-08-14
fix issue #166

josojo:


josojo:
Sorry, first I want to rethink it...

PR #179, opened 2019-09-03
Update gas reporter

bh2smith:
Previous version of gas reporter was not working properly!

**Test Plan:**
See that our continuous integration and units tests are all still fine (since no changes to the code).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,10 +32,40 @@
         "ethereumjs-abi": "^0.6.5"
       }
     },
+    "@types/concat-stream": {
+      "version": "1.6.0",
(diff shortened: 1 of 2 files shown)


fleupold (approved):
What exactly was the issue with the previous version?

bh2smith:
> What exactly was the issue with the previous version?

All I can really get out of it was this error (on my end) 
```
Error: invalid reporter "eth-gas-reporter"
```
which, perhaps, wasn't jiving well with my truffle or node version?

Also @c3rnst was experiencing a different issue that was complaining about a missing `package-lock` file. 

Overall, I suspect this was more of an issue with the fact that we were relying on a global truffle installation rather than the local.

PR #181, opened 2019-09-05
Implements the reward mechanism for the solution submitter

josojo:
This PR grants half of the collected fees to the solution submitter.

I am aware that in the current state a solutionSubmitter could withdraw their reward, right before another solution is submitted. But this will be fixed within another PR dealing with this:
https://github.com/gnosis/dex-contracts/issues/166

Testplan:
unit tests

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -120,7 +120,6 @@ contract StablecoinConverter is EpochTokenLocker {
 
     struct AuctionData {
         int256 feeCollected;
-        address bestSolutionSubmitter;
     }
 
     struct PreviousSolutionData {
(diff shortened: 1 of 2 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -122,6 +122,7 @@ contract StablecoinConverter is EpochTokenLocker {
         uint batchId;
         TradeData[] trades;
         uint16[] tokenIdsForPrice;
+        address solutionSubmitter;
```
fleupold, line 151: Instead of having currentFeeCollected separate, can we store both rewardAmount and rewardReceiver on this struct?

test/stablecoin_converter.js
```diff
@@ -948,6 +948,67 @@ contract("StablecoinConverter", async (accounts) => {
     await stablecoinConverter.submitSolution(batchIndex, owner2, orderIds2, volume2, prices2, tokenIdsForPrice2)
     assert.equal((await stablecoinConverter.currentPrices.call(2)).toNumber(), 0, "CurrentPrice were not adjusted correctly")
   })
+  it("checks that fees are granted correctly", async () => {
```
fleupold: nit: `it("grants fee surplus to solution submitter")`

Maybe also add a test case for the case where there is another solution overriding the first (making sure the original submitter doesn't get also get a reward)?

fleupold (approved):

test/stablecoin_converter.js
josojo: This is exactly the second test, which is added in this PR, wasn't I?

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FxT5LMH1Qi5BisdmMne%2Fgiphy.gif&token=imblind" data-gitmeme-token="imblind"></a>

bh2smith (approved):
This seems alright for the moment, I assume the referenced PR #166 will be merged before this.

PR #201, opened 2019-09-18
rearrage account variable in order_placement.js

josojo:
I always got the error that accounts was not found.

No clue, why it seemed to work for you guys.

---

testplan: place an order with the orderscript


fleupold (approved):
I believe this was a bad merge. It's also addressed in #199

josojo:
Felix found it earlier :)

PR #205, opened 2019-09-19
Inverted buy-sell prices

bh2smith:
This tiny little error somehow snuck in to the contracts. I would hope that the unit tests would fail as a result of this change.

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -185,7 +185,7 @@ contract StablecoinConverter is EpochTokenLocker {
             uint128 executedSellAmount = volumes[i];
             require(currentPrices[order.sellToken] != 0, "prices are not allowed to be zero");
             uint128 executedBuyAmount = getExecutedBuyAmount(
-                volumes[i],
+                executedSellAmount,
                 currentPrices[order.buyToken],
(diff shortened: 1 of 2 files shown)


fleupold (commented):
I'm not sure I like the A, B, C, D stuff. It doesn't show on passing tests and if an assert fails, is it really that useful to see a letter instead of just the error message and the line number? I'd also argue that it's fairly easy for these to get outdated when asserts are added, removed or rearranged.

Did anything have to change apart from flipping the order of the prices in the example (it's hard to tell from the diff since most changes seem to just affect the string which seems unrelated. Maybe we can discuss these changes separately from the actual inversion.

bh2smith:
> just the error message and the line number?

If it showed the line number, I wouldn't have done this. I will check again and revert if this is the case.

josojo (approved):
Now, it is very clean! Sweet deal

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FScjoxvTmUbCW75RtbI%2Fgiphy.gif&token=letsgo" data-gitmeme-token="letsgo"></a>

PR #213, opened 2019-09-25
[Refactor] Trying to improve the readablity of the smart contract

josojo:
---

testplan: unit tests only

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -182,18 +182,9 @@ contract StablecoinConverter is EpochTokenLocker {
         int[] memory tokenConservation = new int[](prices.length);
         for (uint i = 0; i < owners.length; i++) {
             Order memory order = orders[owners[i]][orderIds[i]];
-            require(order.validFrom <= batchIndex, "Order is not yet valid");
-            require(order.validUntil >= batchIndex, "Order is no longer valid");
-            // Assume for now that we always have sellOrders
-            uint128 executedBuyAmount = volumes[i];
-            require(currentPrices[order.sellToken] != 0, "prices are not allowed to be zero");
-            uint128 executedSellAmount = getExecutedSellAmount(
-                executedBuyAmount,
-                currentPrices[order.buyToken],
-                currentPrices[order.sellToken]
-            );
-            tokenConservation[findPriceIndex(order.buyToken, tokenIdsForPrice)] -= int(executedBuyAmount);
-            tokenConservation[findPriceIndex(order.sellToken, tokenIdsForPrice)] += int(executedSellAmount);
+            require(checkOrderValidity(order, batchIndex), "Order is not valid");
+            (uint128 executedBuyAmount, uint128 executedSellAmount) = getTradedAmounts(volumes[i], order);
+            updateTokenConservation(tokenConservation, order, tokenIdsForPrice, executedBuyAmount, executedSellAmount);
             require(order.remainingAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
             // Ensure executed price is not lower than the order price:
             //       executedSellAmount / executedBuyAmount <= order.priceDenominator / order.priceNumerator
@@ -206,11 +197,9 @@ contract StablecoinConverter is EpochTokenLocker {
         }
         // doing all subtractions after all additions (in order to avoid negative values)
         for (uint i = 0; i < owners.length; i++) {
-            uint128 executedBuyAmount = volumes[i];
-            uint128 executedSellAmount = getExecutedSellAmount(
-                executedBuyAmount,
-                currentPrices[orders[owners[i]][orderIds[i]].buyToken],
-                currentPrices[orders[owners[i]][orderIds[i]].sellToken]
+            (uint128 executedBuyAmount, uint128 executedSellAmount) = getTradedAmounts(
+                volumes[i],
+                orders[owners[i]][orderIds[i]]
             );
             subtractBalance(
                 owners[i],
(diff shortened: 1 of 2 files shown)


josojo:
Just to get a feeling of the gas costs increases for this refactoring. 

After refactoring:
```
  ✓ settles a ring trade between 3 tokens (2020ms, 14544757 gas)
submitsolution costs: 773008
```

before:
```
      ✓ settles a ring trade between 3 tokens (1811ms, 14584035 gas)
submitsolution costs: 772192
```
I think it is insignificant and we should keep on doing this kind of refactorings.

contracts/StablecoinConverter.sol
```diff
@@ -347,6 +326,18 @@ contract StablecoinConverter is EpochTokenLocker {
...
+    function updateTokenConservation(
+        int[] memory tokenConservation,
+        Order memory order,
+        uint16[] memory tokenIdsForPrice,
+        uint128 buyAmount,
+        uint128 sellAmount
+    ) private pure returns(int[] memory) {
+        tokenConservation[findPriceIndex(order.buyToken, tokenIdsForPrice)] -= int(buyAmount);
+        tokenConservation[findPriceIndex(order.sellToken, tokenIdsForPrice)] += int(sellAmount);
+        return tokenConservation;
```
fleupold: Is this necessary? Will the array that is passed in not be mutated?

contracts/StablecoinConverter.sol
```diff
@@ -317,26 +306,16 @@ contract StablecoinConverter is EpochTokenLocker {
...
                 uint orderId = previousSolution.trades[i].orderId;
                 Order memory order = orders[owner][orderId];
-                uint128 buyVolume = previousSolution.trades[i].volume;
-                uint128 sellVolume = getExecutedSellAmount(
-                    buyVolume,
-                    currentPrices[order.buyToken],
-                    currentPrices[order.sellToken]
-                );
-                addBalance(owner, tokenIdToAddressMap(order.sellToken), sellVolume);
+                (, uint128 sellAmount) = getTradedAmounts(previousSolution.trades[i].volume, order);
```
fleupold, line 309: One could create a `TradedVolume` struct with two fields and pass this around instead of having these tuples everywhere. 
Not sure if this would be an improvement.

fleupold (approved):
Nice, I like that the unit tests don't have to change at all and we are pretty confident everything still works.

I'm wondering if the TokenConservation logic could live in its own library even.
This library could maintain the data structure and expose the functions `updateTokenConservation`, `updateTokenConservation` and internally implement the `findPriceIndex` method.

It would make the main smart contract smaller and colocate coherent logic.

contracts/StablecoinConverter.sol
```diff
@@ -11,7 +11,7 @@ contract StablecoinConverter is EpochTokenLocker {
     using BytesLib for bytes32;
     using BytesLib for bytes;
 
-    uint constant MAX_UINT128 = 2**128 - 1;
+    uint constant private MAX_UINT128 = 2**128 - 1;
```
bh2smith, line 14: Does making this private improve the readability? Does it save gas?

contracts/StablecoinConverter.sol
```diff
@@ -182,18 +182,9 @@ contract StablecoinConverter is EpochTokenLocker {
         int[] memory tokenConservation = new int[](prices.length);
         for (uint i = 0; i < owners.length; i++) {
             Order memory order = orders[owners[i]][orderIds[i]];
-            require(order.validFrom <= batchIndex, "Order is not yet valid");
-            require(order.validUntil >= batchIndex, "Order is no longer valid");
-            // Assume for now that we always have sellOrders
```
bh2smith, line 187: Has this been removed completely?

contracts/StablecoinConverter.sol
```diff
@@ -206,11 +197,9 @@ contract StablecoinConverter is EpochTokenLocker {
         }
         // doing all subtractions after all additions (in order to avoid negative values)
         for (uint i = 0; i < owners.length; i++) {
-            uint128 executedBuyAmount = volumes[i];
-            uint128 executedSellAmount = getExecutedSellAmount(
-                executedBuyAmount,
-                currentPrices[orders[owners[i]][orderIds[i]].buyToken],
-                currentPrices[orders[owners[i]][orderIds[i]].sellToken]
+            (, uint128 executedSellAmount) = getTradedAmounts(
+                volumes[i],
```
bh2smith, line 201: nit: I prefer the term `amounts` here

contracts/StablecoinConverter.sol
```diff
@@ -347,6 +326,18 @@ contract StablecoinConverter is EpochTokenLocker {
...
     }
 
+    function updateTokenConservation(
+        int[] memory tokenConservation,
+        Order memory order,
+        uint16[] memory tokenIdsForPrice,
+        uint128 buyAmount,
+        uint128 sellAmount
+    ) private pure returns(int[] memory) {
+        tokenConservation[findPriceIndex(order.buyToken, tokenIdsForPrice)] -= int(buyAmount);
```
bh2smith, line 336: I didn't know about `int`... what does it do?

contracts/StablecoinConverter.sol
```diff
@@ -182,18 +182,9 @@ contract StablecoinConverter is EpochTokenLocker {
...
-            uint128 executedBuyAmount = volumes[i];
-            require(currentPrices[order.sellToken] != 0, "prices are not allowed to be zero");
-            uint128 executedSellAmount = getExecutedSellAmount(
-                executedBuyAmount,
-                currentPrices[order.buyToken],
-                currentPrices[order.sellToken]
-            );
-            tokenConservation[findPriceIndex(order.buyToken, tokenIdsForPrice)] -= int(executedBuyAmount);
-            tokenConservation[findPriceIndex(order.sellToken, tokenIdsForPrice)] += int(executedSellAmount);
+            require(checkOrderValidity(order, batchIndex), "Order is not valid");
```
```suggestion
            require(checkOrderValidity(order, batchIndex), "Order is invalid");
```

bh2smith (approved):
Looks nice. Made a couple suggestions!

contracts/StablecoinConverter.sol
fleupold: https://solidity.readthedocs.io/en/v0.5.3/types.html#integers

contracts/StablecoinConverter.sol
josojo: I decided against it, as then we have the very long namings.
`
tradeVolume.buyAmount 
tradeVolume.sellAmount 
`

But yeah, both solutions are possible

contracts/StablecoinConverter.sol
josojo: it is better to always declare the visibility!

contracts/StablecoinConverter.sol
josojo: nope. I even thought of you and made it a todo:

   // ToDo: implement logic also for buyOrders

contracts/StablecoinConverter.sol
josojo: I think volumes is very common. No one talks about trading amounts, it is always trading volume :)

But you are right, that we are inconsistent with volumes and amounts. Might tackle it in another PR

josojo:
> I'm wondering if the TokenConservation logic could live in its own library even.
This library could maintain the data structure and expose the functions updateTokenConservation, updateTokenConservation and internally implement the findPriceIndex method.

I had the same idea! omg. But this is for the next PR.

contracts/StablecoinConverter.sol
```diff
@@ -355,6 +345,22 @@ contract StablecoinConverter is EpochTokenLocker {
         previousSolution.objectiveValue = newObjectiveValue;
     }
 
+    function getTradedAmounts(uint128 volume, Order memory order) private view returns (uint128, uint128) {
+        // ToDo: implement logic also for buyOrders
```
anxolin, line 349: nit: Just fyi, I thin the convention for the ToDos is to follow this pattern:
```js
// TODO: message
```
There's some automatic tools that helps keeping track of them and they don't recognize `ToDo`

image

By the way, I just tried this one and looks nice: https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree

anxolin (commented):
Much more readable! 👍

contracts/StablecoinConverter.sol
josojo: Cool. Good to know

PR #216, opened 2019-10-01
Script modifications for rinkeby

josojo:
I modified the add_token script to receive parameters as all the other scripts, as the old solution would not allow specifying network parameters.

I also added an approve_for_deposit script, as it was needed for me for rinkeby.

---
testplan:
Running the scripts against the Rinkeby deployment


fleupold (commented):
I wonder, should approve for deposit maybe be part of the deposit script?

Also, have you tried running

```
npx truffle exec scripts/stablex/add_token.js --network rinkeby <token address>
```

That should have worked before, too.

fleupold (approved):
I'm still not convinced that adding the extra parameter was necessary but

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FxUOwGcipkmaWjzpVaE%2Fgiphy.gif&token=what_ever" data-gitmeme-token="what_ever"></a>

josojo:
You are right, probably, it was not necessary. Though, now it is handled more uniformly in all scripts.

PR #229, opened 2019-10-15
Objective Function Part II - penalize disregarded utility of touched orders

bh2smith:
This PR is intended to provide an objective function to auction settlement which rewards total traders utility while penalizing solutions whose touched orders have disregarded utility.

In brief, the objective function is `U - dU` where `U`, the total utility, is the sum over all touched orders of their individual trade utilities `u_i`; 
```
 u_i = ((execBuyAmount * sellAmount - execSellAmount * buyAmount) * priceBuyToken) // sellAmount
```
and `dU`, the total disregarded utility, is the sum over all touched orders of their individual disregarded utilities `dU_i`; 
```
dU_i = (((sellAmount - execSellAmount) * priceBuyToken) * buyAmount) // sellAmount
```

The gas reporter gives the following estimates (for a basic trade of two orders).

```
Two Orders:
|···························································································│
|  Optimizer enabled: true  ·         Runs: 200  ·        Block limit: 6721975 gas          │
|···························································································│
|  Methods            ·               20 gwei/gas               ·       183.32 usd/eth      │
|·····················|·················|·········|········|·········|·········|············│
|                     ·  Method         ·  Min    ·  Max   ·  Avg    · # calls ·  usd (avg) │
|  With New Objective ·  submitSolution ·  410182 · 865437 ·  610996 ·  35     · 2.24       │
|  Without Objective  ·  submitSolution ·  406816 · 858665 ·  615099 ·  35     · 2.26       │
|···························································································│
```

TODO - run this same gas estimate on a larger collection of orders!

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -32,7 +32,7 @@ contract StablecoinConverter is EpochTokenLocker {
     struct Order {
         uint16 buyToken;
         uint16 sellToken;
-        uint32 validFrom;  // order is valid from auction collection period: validFrom inclusively
+        uint32 validFrom;   // order is valid from auction collection period: validFrom inclusively
         uint32 validUntil;  // order is valid till auction collection period: validUntil inclusively
         bool isSellOrder;
         uint128 priceNumerator;
@@ -182,10 +182,15 @@ contract StablecoinConverter is EpochTokenLocker {
         updateCurrentPrices(prices, tokenIdsForPrice);
         delete previousSolution.trades;
         int[] memory tokenConservation = new int[](prices.length);
+        int utility = 0;
+        int disregardedUtility = 0;
         for (uint i = 0; i < owners.length; i++) {
             Order memory order = orders[owners[i]][orderIds[i]];
             require(checkOrderValidity(order, batchIndex), "Order is invalid");
             (uint128 executedBuyAmount, uint128 executedSellAmount) = getTradedAmounts(volumes[i], order);
+            // accumulate objective values before updateRemainingOrder
+            utility += evaluateUtility(executedBuyAmount, executedSellAmount, order);
+            disregardedUtility += evaluateDisregardedUtility(executedSellAmount, order);
             updateTokenConservation(tokenConservation, order, tokenIdsForPrice, executedBuyAmount, executedSellAmount);
             require(order.remainingAmount >= executedSellAmount, "executedSellAmount bigger than specified in order");
             // Ensure executed price is not lower than the order price:
@@ -209,15 +214,6 @@ contract StablecoinConverter is EpochTokenLocker {
                 executedSellAmount
             );
         }
-        // Now that all trades have settled (without over/under-flow), we compute objective values.
-        int utility = 0;
-        int disregardedUtility = 0;
-        for (uint i = 0; i < owners.length; i++) {
-            Order memory order = orders[owners[i]][orderIds[i]];
-            (uint128 execBuyAmt, uint128 execSellAmt) = getTradedAmounts(volumes[i], order);
-            utility += evaluateUtility(execBuyAmt, execSellAmt, order);
-            disregardedUtility += evaluateDisrgardedUtility(execSellAmt, order);
-        }
         int objectiveValue = utility - disregardedUtility;
         checkAndOverrideObjectiveValue(objectiveValue);
         grantRewardToSolutionSubmitter(uint(tokenConservation[0]) / 2);
@@ -225,7 +221,7 @@ contract StablecoinConverter is EpochTokenLocker {
         documentTrades(batchIndex, owners, orderIds, volumes, tokenIdsForPrice);
     }
 
-    function getCurrentObjectiveValue() public view returns(uint) {
+    function getCurrentObjectiveValue() public view returns(int) {
         if (previousSolution.batchId == getCurrentBatchId() - 1) {
             return previousSolution.objectiveValue;
         } else {
@@ -233,18 +229,22 @@ contract StablecoinConverter is EpochTokenLocker {
         }
     }
 
-    function evaluateUtility(uint128 execBuy, uint128 execSell, Order memory order) internal returns(int) {
+    function evaluateUtility(uint128 execBuy, uint128 execSell, Order memory order) internal view returns(int) {
         // Utility = ((execBuyAmt * order.sellAmt - execSellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
-        return ((execBuy * order.sellAmount - execSell * order.buyAmount) * currentPrices[order.buyToken]) / order.sellAmount;
+        uint128 buyAmount = order.priceNumerator;
+        uint128 sellAmount = order.priceDenominator;
+        return ((execBuy * sellAmount - execSell * buyAmount) * currentPrices[order.buyToken]) / sellAmount;
     }
 
-    function evaluateDisregardedUtility(uint128 execSell, Order memory order) internal returns(int) {
+    function evaluateDisregardedUtility(uint128 execSell, Order memory order) internal view returns(int) {
         // |disregardedUtility| = maxUtility - actualUtility
-        // where maxUtility is the utility achieved when executedSellAmount == order.sellAmount. That is,
+        // where maxUtility is the utility achieved when execSellAmount == order.sellAmount. That is,
         // maxUtility = ((execBuyAmt * order.sellAmt - order.sellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
         //            = (execBuyAmt - order.buyAmount) * price.buyToken <---- This reduction not used in final formula.
         // So, after simplification
-        return ((execSell - order.sellAmount) * currentPrices[order.buyToken] * order.buyAmount) / order.sellAmount;
+        uint128 buyAmount = order.priceNumerator;
+        uint128 sellAmount = order.priceDenominator;
+        return ((execSell - sellAmount) * currentPrices[order.buyToken] * buyAmount) / sellAmount;
     }
 
     function grantRewardToSolutionSubmitter(uint feeReward) internal {
@@ -358,7 +358,7 @@ contract StablecoinConverter is EpochTokenLocker {
         uint16[] memory tokenIdsForPrice,
         uint128 buyAmount,
         uint128 sellAmount
-    ) private {
+    ) private pure {
         tokenConservation[findPriceIndex(order.buyToken, tokenIdsForPrice)] -= int(buyAmount);
         tokenConservation[findPriceIndex(order.sellToken, tokenIdsForPrice)] += int(sellAmount);
     }
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -54,6 +54,7 @@ contract StablecoinConverter is EpochTokenLocker {
     uint public MAX_TOKENS;  // solhint-disable var-name-mixedcase
     uint16 public numTokens = 0;
     uint128 public feeDenominator; // fee is (1 / feeDenominator)
+    uint128 private utilityAdjustmentFactor = 2;
 
     constructor(uint maxTokens, uint128 _feeDenominator, address feeToken) public {
         MAX_TOKENS = maxTokens;
@@ -235,17 +236,16 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     function evaluateUtility(uint128 execBuy, Order memory order) internal view returns(uint128) {
-        // Utility = ((execBuyAmt * order.sellAmt - scaledSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
-        uint256 scaledSellAmount = getExecutedSellAmount(
+        // Utility = ((execBuyAmt * order.sellAmt - adjustedSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
+        uint256 adjustedSellAmount = getExecutedSellAmount(
             execBuy,
             currentPrices[order.buyToken],
             currentPrices[order.sellToken],
-            2
+            uint128(feeDenominator.mul(utilityAdjustmentFactor))
         );
         return uint128(
-            execBuy.sub(
-                scaledSellAmount.mul(order.priceNumerator).div(order.priceDenominator)
-            ).mul(currentPrices[order.buyToken])
+            execBuy.sub(adjustedSellAmount.mul(order.priceNumerator)
+                .div(order.priceDenominator)).mul(currentPrices[order.buyToken])
         );
     }
 
@@ -270,19 +270,19 @@ contract StablecoinConverter is EpochTokenLocker {
         uint128 executedBuyAmount,
         uint128 buyTokenPrice,
         uint128 sellTokenPrice,
-        uint128 scale  // Used in utility evaluation with scale = 2
+        uint128 feeDenominator
     ) internal view returns (uint128) {
-        // executedSellAmount = sellAmount * (1 - (1/feeDenominator))
-        //                    = sellAmount - sellAmount/feeDenominator
-        //                    = (sellAmount * feeDenominator) / feeDenominator - sellAmount/feeDenominator
-        //                    = (sellAmount * feeDenominator - sellAmount) / feeDenominator
-        //                    = (sellAmount * (feeDenominator - 1)) /feeDenominator
-        //                    = (executedBuyAmount * buyTokenPrice / sellTokenPrice) * (feeDenominator - 1) / feeDenominator
+        // Based on Equation (2) from https://github.com/gnosis/dex-contracts/issues/173#issuecomment-526163117
+        // execSellAmount * p[sellToken] * (1 - phi) == execBuyAmount * p[buyToken]
+        // where phi = 1/feeDenominator
+        // Note that: 1 - phi = (feeDenominator - 1) / feeDenominator
+        // And so, 1/(1-phi) = feeDenominator / (feeDenominator - 1)
+        // executedSellAmount = (execBuyAmount * p[buyToken]) / (p[sellToken] * (1 - phi))
+        //                    = (execBuyAmount * buyTokenPrice / sellTokenPrice) * feeDenominator / (feeDenominator - 1)
         //                    in order to minimize rounding errors, the order is switched
-        //                    = (executedBuyAmount * buyTokenPrice / feeDenominator) * (feeDenominator - 1) / sellTokenPrice
-        uint128 scaledFeeDenominator = scale * feeDenominator;
-        uint256 sellAmount = (uint256(executedBuyAmount).mul(buyTokenPrice) / (scaledFeeDenominator - 1))
-            .mul(scaledFeeDenominator) / sellTokenPrice;
+        //                    = ((executedBuyAmount * buyTokenPrice) / (feeDenominator - 1)) * feeDenominator) / sellTokenPrice
+        uint256 sellAmount = uint256(executedBuyAmount).mul(buyTokenPrice).div(feeDenominator - 1)
+            .mul(feeDenominator).div(sellTokenPrice);
         require(sellAmount < MAX_UINT128, "sellAmount too large");
         return uint128(sellAmount);
     }
@@ -364,7 +364,7 @@ contract StablecoinConverter is EpochTokenLocker {
             executedBuyAmount,
             currentPrices[order.buyToken],
             currentPrices[order.sellToken],
-            1
+            feeDenominator
         );
         return (executedBuyAmount, executedSellAmount);
     }
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -235,14 +235,13 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     function evaluateUtility(uint128 execBuy, Order memory order) internal view returns(uint128) {
-        // Utility = ((execBuyAmt * order.sellAmt - preFeeSell * order.buyAmt) * price.buyToken) / order.sellAmt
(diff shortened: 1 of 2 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,46 @@ contract StablecoinConverter is EpochTokenLocker {
...
         } else {
-            return 0;
+            return MIN_INT;
         }
     }
 
+    function evaluateUtility(uint128 execBuy, uint128 execSell, Order memory order) internal view returns(int) {
+        // Utility = ((execBuyAmt * order.sellAmt - execSellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
+        (uint128 buyAmount, uint128 sellAmount) = getBuySellAmounts(order);
+        return ((execBuy * sellAmount - execSell * buyAmount) * currentPrices[order.buyToken]) / sellAmount;
```
fleupold: We need to watch out for overflow here. Should probably convert amounts to u256 and try to keep it to two multiplications (divide before multiplying with current prices) to make sure we don't overflow.

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,46 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        // |disregardedUtility| = maxUtility - actualUtility
+        // where maxUtility is the utility achieved when execSellAmount == order.sellAmount. That is,
+        // maxUtility = ((execBuyAmt * order.sellAmt - order.sellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
+        //            = (execBuyAmt - order.buyAmount) * price.buyToken <---- This reduction not used in final formula.
+        // So, after simplification
+        (uint128 buyAmount, uint128 sellAmount) = getBuySellAmounts(order);
+        return (((sellAmount - execSell) * currentPrices[order.buyToken]) * buyAmount) / sellAmount;
+    }
+
+    function getBuySellAmounts(Order memory order) internal pure returns(uint128, uint128) {
```
fleupold: nit BuyAndSellAmounts?

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,46 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        (uint128 buyAmount, uint128 sellAmount) = getBuySellAmounts(order);
+        return ((execBuy * sellAmount - execSell * buyAmount) * currentPrices[order.buyToken]) / sellAmount;
+    }
+
+    function evaluateDisregardedUtility(uint128 execSell, Order memory order) internal view returns(int) {
+        // |disregardedUtility| = maxUtility - actualUtility
+        // where maxUtility is the utility achieved when execSellAmount == order.sellAmount. That is,
+        // maxUtility = ((execBuyAmt * order.sellAmt - order.sellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
+        //            = (execBuyAmt - order.buyAmount) * price.buyToken <---- This reduction not used in final formula.
+        // So, after simplification
```
fleupold: Does it make sense to show the simplification here? It's not super obvious why execBuyAmt is no input to this function, but I guess the reader can also do the math himself?

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,46 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        (uint128 buyAmount, uint128 sellAmount) = getBuySellAmounts(order);
+        return (((sellAmount - execSell) * currentPrices[order.buyToken]) * buyAmount) / sellAmount;
```
fleupold: Does this take into account if the account's balance would have been able to afford this trade? I think this is a crucial non-trivial requirement for that metric.

fleupold (commented):
Cool stuff. Some more unit tests would be great to document how this metric behaves in certain cases (e.g. when disregarded orders are not touched or if the account doesn't have enough balance).

contracts/StablecoinConverter.sol
bh2smith: Yes, I had this on my mind but wanted to "get something out there" for us to look at and evaluate beforehand. Not that I have already experimented with reordering the last division with the last multiplication like this: 

```py
((execBuy * sellAmt - execSell * buyAmt) / sellAmt) * price[buyToken]
```
It returns 0 in undesired places (such as our basic trade unit test, for example). The reason for this is that the magnitude of the difference obtained in the first term is usually not very large. So dividing immediately will often result in 0. My  best approach for recovering from this is to **distribute the arithmetic inside the brackets and take the difference at the very end**. Ya, that will probably work! So, new formula will be...

```py
execBuy * price[buyToken] - ((execSell * buyAmt) / sellAmt) * price[buyToken]
```

or possibly, one step simplified, 

```py
(execBuy - (execSell * buyAmt) / sellAmt) * price[buyToken]
```

contracts/StablecoinConverter.sol
bh2smith: This was just included for my reviewers. Will disappear before merge!

contracts/StablecoinConverter.sol
bh2smith: I believe it does... since `submitSolution` would fail if it didn't.

contracts/StablecoinConverter.sol
bh2smith: Ya, so that is nice, but the next equation (for disregarded utility) is going to be a bit more difficult to rearrange in a meaningful way without losing information...

```
(((sellAmt - execSell) * price[buyToken]) * buyAmt) / sellAmt;
```

Perhaps its best if I evaluate `maxUtility` with the utility formula above and simply take the difference 

```
dU = maxU - U
```

nlordell: Don't know if this helps but could we try using [fixed point numbers](https://solidity.readthedocs.io/en/latest/types.html#fixed-point-numbers)? It looks like its not fully supported yet but maybe enough for us to use? We are operating on 128bit integers we are wasting 128 bits of precision (since, if I understand correctly, all math ops are done on 256 bit integers anyway). We could use fixed196x64 which gives us 64 bits of decimal precision and 64 extra bits to handle some overflow in the multiplications and additions. It may provide more accurate results.

bh2smith: This is a cool idea, although I found the following warning at the top of the description:

```
Fixed point numbers are not fully supported by Solidity yet. 
They can be declared, but cannot be assigned to or from.
```
I'll keep this open for now and experiment in a separate sandbox!

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,43 @@ contract StablecoinConverter is EpochTokenLocker {
                 executedSellAmount
             );
         }
-        checkAndOverrideObjectiveValue(uint(tokenConservation[0]));
+        int objectiveValue = utility - disregardedUtility;
```
josojo: Intuitiviely, I would go for uint instead of int for the objective value.

Because:
1. I can always submit the trivial solution without orders, which will have an objective value of 0
2. Any valuable solution in the 2-dimensional case has a positive utility- disregarded utility.
3. overflow issue handling with ints is much harder.

But for sure, we should check it more in detail

contracts/StablecoinConverter.sol
```diff
@@ -182,10 +183,15 @@ contract StablecoinConverter is EpochTokenLocker {
...
         delete previousSolution.trades;
         int[] memory tokenConservation = new int[](prices.length);
+        int utility = 0;
+        int disregardedUtility = 0;
         for (uint i = 0; i < owners.length; i++) {
             Order memory order = orders[owners[i]][orderIds[i]];
             require(checkOrderValidity(order, batchIndex), "Order is invalid");
             (uint128 executedBuyAmount, uint128 executedSellAmount) = getTradedAmounts(volumes[i], order);
+            // accumulate objective values before updateRemainingOrder
+            utility += evaluateUtility(executedBuyAmount, executedSellAmount, order);
```
josojo: please use the safe math operations

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,43 @@ contract StablecoinConverter is EpochTokenLocker {
...
         documentTrades(batchIndex, owners, orderIds, volumes, tokenIdsForPrice);
     }
 
-    function getCurrentObjectiveValue() public view returns(uint) {
+    function getCurrentObjectiveValue() public view returns(int) {
         if (previousSolution.batchId == getCurrentBatchId() - 1) {
             return previousSolution.objectiveValue;
         } else {
-            return 0;
+            return MIN_INT;
```
josojo: because of the previous comment, I guess we can keep 0.

But I am not sure

contracts/StablecoinConverter.sol
```diff
@@ -209,20 +215,43 @@ contract StablecoinConverter is EpochTokenLocker {
...
         }
     }
 
+    function evaluateUtility(uint128 execBuy, uint128 execSell, Order memory order) internal view returns(int) {
+        // Utility = ((execBuyAmt * order.sellAmt - execSellAmt * order.buyAmt) * price.buyToken) / order.sellAmt
+        (uint128 buyAmount, uint128 sellAmount) = getBuyAndSellAmounts(order);
+        return (execBuy - ((execSell * buyAmount) / sellAmount)) * currentPrices[order.buyToken];
```
josojo: Why do you favor this solution over this one_

        return ((execBuy * sellAmount - execSell * buyAmount) / sellAmount)) * currentPrices[order.buyToken];

I don't know which one is better... Just asking stupid questions here :)

And please use safe math

contracts/StablecoinConverter.sol
bh2smith: Right, yes the trivial solution is considered "better" than other non-trivial ones with a lot of disregarded utility. This, kind-of, implies nobody would submit such a solution... because someone else would immediately submit the trivial solution which is better. In this case I would impose the following requirement

```js
require(utility >= disregardedUtility, "solution must be better than trivial");
```

contracts/StablecoinConverter.sol
bh2smith: yes good idea! Will do.

contracts/StablecoinConverter.sol
bh2smith: Right cool!

contracts/StablecoinConverter.sol
bh2smith: This alternative solution does not work because of the following example:
suppose `(a, b, c, d) = (1, 2, 3, 3)`
```
(a * b * c) / d = 6 / 3 = 2
BUT
((a * b) / d) * c = 0 * 3 = 0
```

Essentially, dividing first results in an undesired loss of precision. I have been working secretly on a Rational Numbers library in Solidity that might help with this stuff, but its not looking promising. Will bring it up tomorrow in the Smart Contracts BiWeekly

contracts/StablecoinConverter.sol
josojo: I believe it does not. All checks are done with regards to the volume, submitted as part of the solution if I recall correctly. But your function "just" looks at the order.remainingAmount

test/stablecoin_converter.js
```diff
@@ -190,6 +190,7 @@ contract("StablecoinConverter", async (accounts) => {
       assert.equal((await stablecoinConverter.getBalance.call(user_1, erc20_2.address)).toNumber(), volume[0], "Bought tokens were not adjusted correctly")
       assert.equal((await stablecoinConverter.getBalance.call(user_2, erc20_2.address)).toNumber(), basicTrade.deposits[1].amount - getSellVolume(volume[1], prices[1], prices[0]), "Sold tokens were not adjusted correctly")
       assert.equal((await stablecoinConverter.getBalance.call(user_2, feeToken.address)).toNumber(), volume[1], "Bought tokens were not adjusted correctly")
+      // None of these assertions implies utility == 0?
```
josojo: Good observation. 

Now, you could check whether it is actually true.

test/stablecoin_converter.js
```diff
@@ -1060,21 +1056,41 @@ contract("StablecoinConverter", async (accounts) => {
...
+        evaluateTradeUtility(basicTrade.orders[1].buyAmount, basicTrade.orders[1].sellAmount, 7992, 4000, prices[0])
+      ]
+      const totalUtility = tradeUtilities.reduce((a, b) => a + b, 0)
+      const disregardedUtilites = [
+        disregardedUtility(basicTrade.orders[0].buyAmount, basicTrade.orders[0].sellAmount, 4000, 8000, prices[1]),
+        disregardedUtility(basicTrade.orders[1].buyAmount, basicTrade.orders[1].sellAmount, 7992, 4000, prices[0])
+      ]
+      const totalDisregardedUtility = disregardedUtilites.reduce((a, b) => a + b, 0)
+      const actualObjectiveValue = (await stablecoinConverter.getCurrentObjectiveValue.call()).toNumber()
+      assert.equal(actualObjectiveValue, totalUtility - totalDisregardedUtility, "Objective value is not stored correct")
```
josojo: really minor: usually, the expected value is the first value

test/stablecoin_converter.js
```diff
@@ -1060,21 +1056,41 @@ contract("StablecoinConverter", async (accounts) => {
...
       const orderId = [orderId1, orderId2]
       const volume = [4000, feeSubtracted(4000) * 2]
       const tokenIdsForPrice = basicTrade.solution.tokenIdsForPrice
 
       await stablecoinConverter.submitSolution(batchIndex, owner, orderId, volume, prices, tokenIdsForPrice, { from: solutionSubmitter })
-
-      assert.equal((await stablecoinConverter.getCurrentObjectiveValue.call()).toNumber(), (getSellVolume(volume[0], prices[0], prices[1]) - volume[1]), "Objective value is not stored correct")
+      const tradeUtilities = [
+        evaluateTradeUtility(basicTrade.orders[0].buyAmount, basicTrade.orders[0].sellAmount, 4000, 8000, prices[1]),
+        evaluateTradeUtility(basicTrade.orders[1].buyAmount, basicTrade.orders[1].sellAmount, 7992, 4000, prices[0])
```
josojo: Could you maybe derive the values for the executed buyAmount and executedSellamout directly from the volume variables? I don't understand, why for the order[0], there is no fee subtraction happing, but for the order[1], there is.

josojo:
scraped

josojo:
This PR means great progress. I love it. You are pushing it @bh2smith!

contracts/StablecoinConverter.sol
bh2smith: and `order.remainingAmount` should be up to date with the balance. Unless it is not updated whenever deposits and withdraws are made/requested.

contracts/StablecoinConverter.sol
bh2smith: This is unfortunately not going to work as expected, since some of our unit tests have fairly large negative objective values.

contracts/StablecoinConverter.sol
bh2smith: If this is not the case, I would argue that this is an unrelated issue to the content of the PR.

test/stablecoin_converter.js
bh2smith: I believe exactly the opposite is true.

Screenshot 2019-10-22 at 10 20 27 AM

Screenshot 2019-10-22 at 10 20 49 AM

bh2smith:
@josojo 
> dU_i = (((sellAmount - execSellAmount) * priceSellToken)

I like this idea because it would be a much simpler calculation. It probably wouldn't hurt to provide some justification for the equation or at least some way to move between our definition here and this one.

In other words, can we verify that
```py
(priceBuyToken * buyAmount) // sellAmount == priceSellToken
```

contracts/StablecoinConverter.sol
```diff
@@ -156,7 +156,7 @@ contract StablecoinConverter is EpochTokenLocker {
         uint16[] tokenIdsForPrice;
         address solutionSubmitter;
         uint256 feeReward;
-        uint256 objectiveValue;
+        uint objectiveValue;
```
josojo, line 157: I liked uint256 better :)

josojo (approved):
Really great!

Thanks for splitting the PRs. Much easier to review.

I think it is also great, that you do the normalization of the prices in the other PR.

contracts/StablecoinConverter.sol
fleupold: Shouldn't this be an int? Are we sure disregarded utility can never be larger than utility?

bh2smith: I started with that, but the argument (which must have been offline) went as follows, the unsafty introduced with int is not worth the nuisance and since the trivial solution evaluates to zero, there is no point in having anything which evaluates below. 

Also notice the require statement below which enforces `utility > disregardedUtility`.

Until now, I have only managed to construct truly crappy, trivial,  examples whose objective is negative. For example, using the basicTrade consisting of two orders from the unit tests. One must fulfill 70% of each order\s requested buy amount in to meet the positivity threshold from our conditions. This suggests that fulfilment of these orders below 70% is considered a "worse solution" by our metric. 

Of course this is not exactly the ideal scenario for our objective function, but it does appear to capture quite a bit.

contracts/StablecoinConverter.sol
```diff
@@ -244,6 +251,22 @@ contract StablecoinConverter is EpochTokenLocker {
         );
     }
 
+    function evaluateDisregardedUtility(Order memory order, address user) internal view returns(uint128) {
+        // |disregardedUtility| = maxUtility - actualUtility
```
fleupold: If this is is the definition, shouldn't we just call `evaluateUtility(remainingBuyAmount, order) - evaluateUtility(executedBuyAmount, order)`?

bh2smith: Not quite, since utility involves a slightly different version of executedSellAmount and this is also no longer the formula being used. Thanks for pointing this out, I will update it!

contracts/StablecoinConverter.sol
fleupold: I'm convinced it's ok to be a unsigned integer, but I also agree that it should make the width explicit and thus stay an uint**256**

contracts/StablecoinConverter.sol
```diff
@@ -249,6 +255,22 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        // Balances and orders have all been updated so: sellAmount - execSellAmt == order.remainingAmount.
+        // For security reasons, we take the minimum of this with the user's token balance.
+        uint256 leftoverSellAmount = Math.min(
+            uint256(order.remainingAmount),
+            getBalance(user, tokenIdToAddressMap(order.sellToken))
+        );
+        // TODO - use SafeCast
+        uint256 limitTerm = currentPrices[order.sellToken].mul(order.priceDenominator)
+            .sub(currentPrices[order.buyToken].mul(order.priceNumerator));
+        return uint128(leftoverSellAmount.mul(limitTerm)) / order.priceDenominator;
```
fleupold: nit: SafeMath and shouldn't the cast happen after the division?

contracts/StablecoinConverter.sol
```diff
@@ -249,6 +255,22 @@ contract StablecoinConverter is EpochTokenLocker {
         );
     }
 
+    function evaluateDisregardedUtility(Order memory order, address user) internal view returns(uint128) {
+        // |disregardedUtility| = (limitTerm * leftoverSellAmount) / order.sellAmount
+        // where limitTerm = price.SellToken * order.sellAmt - order.buyAmt * price.buyToken
+        // and leftoverSellAmount = order.sellAmt - execSellAmt
+        // Balances and orders have all been updated so: sellAmount - execSellAmt == order.remainingAmount.
+        // For security reasons, we take the minimum of this with the user's token balance.
```
fleupold: It's not really a security measure but just correctness, right?

contracts/StablecoinConverter.sol
bh2smith: I agree that `int` is more natural to consider in this scenario, but the argument about the trivial solution is pretty strong. Glad you are happy to use `uint`.

contracts/StablecoinConverter.sol
bh2smith: Yes, I suppose so!

test/stablecoin_converter.js
```diff
@@ -1137,7 +1148,40 @@ contract("StablecoinConverter", async (accounts) => {
         "sellAmount too large"
       )
     })
+    it("reverts if max touched orders is exceeded", async () => {
```
fleupold: This should have probably been a separate PR?

contracts/StablecoinConverter.sol
```diff
@@ -249,6 +255,22 @@ contract StablecoinConverter is EpochTokenLocker {
         );
     }
 
+    function evaluateDisregardedUtility(Order memory order, address user) internal view returns(uint128) {
+        // |disregardedUtility| = (limitTerm * leftoverSellAmount) / order.sellAmount
+        // where limitTerm = price.SellToken * order.sellAmt - order.buyAmt * price.buyToken
```
fleupold, line 257: Could you maybe explain what this "limitTerm" is supposed to represent?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FiXQ8SgaMQAgtq%2Fgiphy.gif&token=lgtm" data-gitmeme-token="lgtm"></a>

test/stablecoin_converter.js
bh2smith: Do we not expect max touched orders to already have been included before this is merged?

contracts/StablecoinConverter.sol
bh2smith: Yes certainly! But only after I fix the failing utility evaluation test!

test/stablecoin_converter.js
bh2smith: Ya... #232  have already been merged... oh ya... weird... I suppose this should already be in the merged PR...

PR #234, opened 2019-10-23
Extracting tokenConservation check into own library

josojo:
as title. This is a refactoring suggested by Felix quite some time ago.

Migration scripts do not need to be adapted, as we just have internal library calls.

---
testplan: 

just unit tests


bh2smith (approved):
This is a very cool change! I really like it a lot. Thanks for moving this! One word "Kickass"!. I would also suggest (maybe) that we also offload the objective function evaluation and comparison to a separate library at some point!



> This is a refactoring suggested by Felix quite some time ago.

Where did this suggestion occur? Can we recover the text? I think it would be best (when referring to other comments) if we can provide references to/for this kinda stuff.

josojo:
Felix proposed it over here: 
https://github.com/gnosis/dex-contracts/pull/213

PR #235, opened 2019-10-24
Signed safe math for our tokenBalance checking

josojo:
- Introduces SignedSafeMath for the token conservation checking
-  updates on missing safeMath operation

The result should be a smart contract that will not allow any overflow operation.

The guys from openzepplin confirmed that the signedSafeMath library is safe and audited, although it is still in draft.

--- 
unit tests


bh2smith (approved):
Cool beans! 🍃

PR #236, opened 2019-10-25
Objective Fuction Part I - Total Utility as objective.

bh2smith:
This is the first of a two part PR. We introduce our Exchange "solution" objective function as the total utility achieved by order matching. This metric is a token independent measure of the trader's utility obtained by getting a better price than specified in the order. Heuristically, the utility of an executed order is 

```
Utility = ((execBuyAmt * order.sellAmt - adjustedSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
```

Note that the naive formula would evaluate to zero when the trader gets exactly their specified price. In order to nudge this above zero, the utility is calculated with `executedSellAmt` replaced by an adjusted version `adjustedSellAmount` which is computed with an adjustment factor on the feeDenominator. 

This is based on Equation (2) from https://github.com/gnosis/dex-contracts/issues/173#issuecomment-526163117 with the fee `phi` replaced by `phi/2`
That is, 

the usual executedSellAmount is defined by this relation,
```
execSellAmount * p[sellToken] * (1 - phi) == execBuyAmount * p[buyToken]
```
while the `adjustedSellAmount` is defined by using half the fee as
```
adjustedSellAmount * p[sellToken] * (1 - phi/2) == execBuyAmount * p[buyToken]
```

This "nudge" serves two purposes:

1. Circumvent the many solutions that would be ranked equally amongst trivial solution (having zero utility) and thus, the smart contract would reject perfectly valid non-trivial solutions.
2. To ensure that direct trades are preferred over long rings which try to exploit market orders.

Finally, be aware that some of the unit tests have been shifted (ever so slightly) to accommodate for the second part of this PR. I would be willing to revert these changes if it makes the understanding more clear.

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -54,6 +54,7 @@ contract StablecoinConverter is EpochTokenLocker {
     uint public MAX_TOKENS;  // solhint-disable var-name-mixedcase
     uint16 public numTokens = 0;
     uint128 public feeDenominator; // fee is (1 / feeDenominator)
+    uint128 private utilityAdjustmentFactor = 2;
 
     constructor(uint maxTokens, uint128 _feeDenominator, address feeToken) public {
         MAX_TOKENS = maxTokens;
@@ -235,17 +236,16 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     function evaluateUtility(uint128 execBuy, Order memory order) internal view returns(uint128) {
-        // Utility = ((execBuyAmt * order.sellAmt - scaledSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
-        uint256 scaledSellAmount = getExecutedSellAmount(
+        // Utility = ((execBuyAmt * order.sellAmt - adjustedSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
+        uint256 adjustedSellAmount = getExecutedSellAmount(
             execBuy,
             currentPrices[order.buyToken],
             currentPrices[order.sellToken],
-            2
+            uint128(feeDenominator.mul(utilityAdjustmentFactor))
         );
         return uint128(
-            execBuy.sub(
-                scaledSellAmount.mul(order.priceNumerator).div(order.priceDenominator)
-            ).mul(currentPrices[order.buyToken])
+            execBuy.sub(adjustedSellAmount.mul(order.priceNumerator)
+                .div(order.priceDenominator)).mul(currentPrices[order.buyToken])
         );
     }
 
@@ -270,19 +270,19 @@ contract StablecoinConverter is EpochTokenLocker {
         uint128 executedBuyAmount,
         uint128 buyTokenPrice,
         uint128 sellTokenPrice,
-        uint128 scale  // Used in utility evaluation with scale = 2
+        uint128 feeDenominator
     ) internal view returns (uint128) {
-        // executedSellAmount = sellAmount * (1 - (1/feeDenominator))
-        //                    = sellAmount - sellAmount/feeDenominator
-        //                    = (sellAmount * feeDenominator) / feeDenominator - sellAmount/feeDenominator
-        //                    = (sellAmount * feeDenominator - sellAmount) / feeDenominator
-        //                    = (sellAmount * (feeDenominator - 1)) /feeDenominator
-        //                    = (executedBuyAmount * buyTokenPrice / sellTokenPrice) * (feeDenominator - 1) / feeDenominator
+        // Based on Equation (2) from https://github.com/gnosis/dex-contracts/issues/173#issuecomment-526163117
+        // execSellAmount * p[sellToken] * (1 - phi) == execBuyAmount * p[buyToken]
+        // where phi = 1/feeDenominator
+        // Note that: 1 - phi = (feeDenominator - 1) / feeDenominator
+        // And so, 1/(1-phi) = feeDenominator / (feeDenominator - 1)
+        // executedSellAmount = (execBuyAmount * p[buyToken]) / (p[sellToken] * (1 - phi))
+        //                    = (execBuyAmount * buyTokenPrice / sellTokenPrice) * feeDenominator / (feeDenominator - 1)
         //                    in order to minimize rounding errors, the order is switched
-        //                    = (executedBuyAmount * buyTokenPrice / feeDenominator) * (feeDenominator - 1) / sellTokenPrice
-        uint128 scaledFeeDenominator = scale * feeDenominator;
-        uint256 sellAmount = (uint256(executedBuyAmount).mul(buyTokenPrice) / (scaledFeeDenominator - 1))
-            .mul(scaledFeeDenominator) / sellTokenPrice;
+        //                    = ((executedBuyAmount * buyTokenPrice) / (feeDenominator - 1)) * feeDenominator) / sellTokenPrice
+        uint256 sellAmount = uint256(executedBuyAmount).mul(buyTokenPrice).div(feeDenominator - 1)
+            .mul(feeDenominator).div(sellTokenPrice);
         require(sellAmount < MAX_UINT128, "sellAmount too large");
         return uint128(sellAmount);
     }
@@ -364,7 +364,7 @@ contract StablecoinConverter is EpochTokenLocker {
             executedBuyAmount,
             currentPrices[order.buyToken],
             currentPrices[order.sellToken],
-            1
+            feeDenominator
         );
         return (executedBuyAmount, executedSellAmount);
     }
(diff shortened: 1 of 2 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -231,6 +234,12 @@ contract StablecoinConverter is EpochTokenLocker {
         }
     }
 
+    function evaluateUtility(uint128 execBuy, Order memory order) internal view returns(uint128) {
+        // Utility = ((execBuyAmt * order.sellAmt - preFeeSell * order.buyAmt) * price.buyToken) / order.sellAmt
+        uint128 preFeeSell = uint128(execBuy.mul(currentPrices[order.buyToken]).div(currentPrices[order.sellToken]));
```
josojo: I would use uint256 here, as in the next line everything will be computed with uin256 anyways, if we use safemath

contracts/StablecoinConverter.sol
```diff
@@ -231,6 +234,12 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        return (execBuy - ((preFeeSell * order.priceNumerator) / order.priceDenominator)) * currentPrices[order.buyToken];
```
josojo: please safemath it

test/stablecoin_converter.js
```diff
@@ -433,12 +432,13 @@ contract("StablecoinConverter", async (accounts) => {
...
 
-      assert.equal((await stablecoinConverter.getBalance.call(user_1, feeToken.address)).toNumber(), basicTrade.deposits[0].amount - 2 * getSellVolume(volume[0], prices[0], prices[1]), "Sold tokens were not adjusted correctly")
-      assert.equal((await stablecoinConverter.getBalance.call(user_1, erc20_2.address)).toNumber(), 2 * volume[0], "Bought tokens were not adjusted correctly")
-      assert.equal((await stablecoinConverter.getBalance.call(user_2, erc20_2.address)).toNumber(), basicTrade.deposits[1].amount - 2 * getSellVolume(volume[1], prices[1], prices[0]), "Sold tokens were not adjusted correctly")
-      assert.equal((await stablecoinConverter.getBalance.call(user_2, feeToken.address)).toNumber(), 2 * volume[1], "Bought tokens were not adjusted correctly")
+      assert.equal((await stablecoinConverter.getBalance.call(user_1, feeToken.address)).toNumber(), basicTrade.deposits[0].amount - getSellVolume(volume[0], prices[0], prices[1]) - getSellVolume(volume2[0], prices[0], prices[1]), "Sold tokens were not adjusted correctly")
```
josojo: getSellVolume(volume[0]+volume2[0], prices[0], prices[1]) would be a little bit more succinct compared to these two function calls

josojo (approved):
Very clean PR. 

I really like this solution for the "always positive utiltity!" 

I checked the tests, they are still testing the same.

Can't wait for the second PR :)

test/stablecoin_converter.js
bh2smith: Yes, I can do that, but this wouldn't hold if we had used different prices for the second solution submission.

bh2smith: Also, I am not entirely sure this is true since `getSellVolume` uses integer division.  Perhaps in this example it would be true, but I think I can construct valid examples that would violate this claim without having different prices. 

```js

function getSellVolume(x, priceNumerator, priceDenominator) {
  return Math.floor(Math.floor(x * priceDenominator / (feeDenominator - 1)) * feeDenominator / priceNumerator)
}
```

contracts/StablecoinConverter.sol
```diff
@@ -231,6 +234,21 @@ contract StablecoinConverter is EpochTokenLocker {
         }
     }
 
+    function evaluateUtility(uint128 execBuy, Order memory order) internal view returns(uint128) {
+        // Utility = ((execBuyAmt * order.sellAmt - scaledSellAmount * order.buyAmt) * price.buyToken) / order.sellAmt
+        uint256 scaledSellAmount = getExecutedSellAmount(
```
fleupold: I'm not a huge fan of the term "scaled sell amount". It reminds me of taking the sellAmount and scaling it up or down. What we really scale here is how much fee is deducted.

Maybe `getExecutedSellAmount` should take `fee Denominator` instead of `scale` as an argument. In the regular case we use the original denominator, and for utility calculation we scale the fee beforehand.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fmgm3FyujKdSzm%2Fgiphy.gif&token=utility" data-gitmeme-token="utility"></a>

Just one small comment about naming, to make it more clear for outsiders what's happening. Couldn't hurt to add a comment on why we are taking half of the fee off for utility calculation.

contracts/StablecoinConverter.sol
```diff
@@ -261,8 +280,9 @@ contract StablecoinConverter is EpochTokenLocker {
         //                    = (executedBuyAmount * buyTokenPrice / sellTokenPrice) * (feeDenominator - 1) / feeDenominator
         //                    in order to minimize rounding errors, the order is switched
         //                    = (executedBuyAmount * buyTokenPrice / feeDenominator) * (feeDenominator - 1) / sellTokenPrice
-        uint256 sellAmount = (uint256(executedBuyAmount).mul(buyTokenPrice) / (feeDenominator - 1))
-            .mul(feeDenominator) / sellTokenPrice;
+        uint128 scaledFeeDenominator = scale * feeDenominator;
```
josojo: safeMath please.

contracts/StablecoinConverter.sol
```diff
@@ -261,8 +280,9 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint256 sellAmount = (uint256(executedBuyAmount).mul(buyTokenPrice) / (scaledFeeDenominator - 1))
```
josojo: IN the comments, it says that (feeDenominator - 1) should be the numerator.
But you are using (scaledFeeDenominator - 1) as a denominator. What is true?

josojo (approved):
Excited to get that one merged!

Great PR, as I said before!

contracts/StablecoinConverter.sol
bh2smith: Funny! We had spoken yesterday about using SafeMath here, but decided it wasn't necessary since this is an internal function that is only ever called by two constant values  (`scale = 1 or 2` and `feeDenominator`). Do you still think we need SafeMath?

contracts/StablecoinConverter.sol
bh2smith: Interesting... I am not sure what happened here. This is a really good catch! For reference, I will include the comments above and adjust this immediately!

```
        // executedSellAmount = sellAmount * (1 - (1/feeDenominator))
        //                    = sellAmount - sellAmount/feeDenominator
        //                    = (sellAmount * feeDenominator) / feeDenominator - sellAmount/feeDenominator
        //                    = (sellAmount * feeDenominator - sellAmount) / feeDenominator
        //                    = (sellAmount * (feeDenominator - 1)) /feeDenominator
        //                    = (executedBuyAmount * buyTokenPrice / sellTokenPrice) * (feeDenominator - 1) / feeDenominator
        //                    in order to minimize rounding errors, the order is switched
        //                    = (executedBuyAmount * buyTokenPrice / feeDenominator) * (feeDenominator - 1) / sellTokenPrice
```

bh2smith: Also, we should be using `SafeMath.div` here.

contracts/StablecoinConverter.sol
bh2smith: Yes, I could do this!

contracts/StablecoinConverter.sol
josojo: If I am placing an order with feeDenominator = maxUint128 - 1, it will overflow.
Not sure whether it can be used for an attack. Thinking this through is too cumbersome, so we should just use mathSafe

contracts/StablecoinConverter.sol
fleupold: Can we add a test case that would have revealed this bug?

contracts/StablecoinConverter.sol
bh2smith: `feeDenominator` is a constant, but it is a parameter of the constructor.. so I will just use `SafeMath`. Better Safe than sorry!

contracts/StablecoinConverter.sol
bh2smith: This is actually not a bug. it is just that the commented equation here was outdated! Nothing to see here boys.

contracts/StablecoinConverter.sol
```diff
@@ -248,17 +266,20 @@ contract StablecoinConverter is EpochTokenLocker {
         uint128 executedBuyAmount,
         uint128 buyTokenPrice,
         uint128 sellTokenPrice
+        // uint128 feeDenominator
```
fleupold, line 269: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FFGTVmzksb2j0k%2Fgiphy.gif&token=oops" data-gitmeme-token="oops"></a>

bh2smith: Damn it!

PR #240, opened 2019-10-28
Price Scaling Hack

bh2smith:
At the moment, the smart contract accepts `prices = [1, 1]` and `prices = [10, 10]` with identical trading volume as having `objectiveValue = 10 and 100` respectively.  Note that this is coming from the fact that utility is the sum of terms each having `priceBuyToken` as a factor.

This is not ideal and could be avoided with something like the following.

```js
require(extended_gcd(prices) ==1, "Must provide prices in reduced form");
```

I have implemented a `solidity_gcd` for element pairs, but will need to evaluate the "exentended_gcd" as follows

```js
pragma solidity ^0.5.0;

library NumberTheory {
    function gcd(uint a, uint b) public pure returns (uint) {
        if (b % a == 0) {
            return a;
        }
        return gcd(b % a, a);
    }

    function extended_gcd(uint[] memory arr) public pure returns (uint) {
        require(arr.length > 0, "can't compute gcd of single element");
        if (arr.length == 1) {
            return arr[0];
        }
        uint res = arr[0];
        for (uint i = 1; i < arr.length; i++) {
            res = gcd(res, arr[i]);
        }
        return res;
    }
}
```

Could easily have O(1) popping if `prices` were a linked list, but this would involve more involved data structures. 

_Keywords:_ Solidity GCD, Solidity Number Theory

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -64,17 +64,17 @@ contract EpochTokenLocker {
     function requestWithdraw(address token, uint amount) public {
         // first process old pendingWithdraw, as otherwise balances might increase for currentBatchId - 1
         if (hasValidWithdrawRequest(msg.sender, token)) {
-            withdraw(token, msg.sender);
+            withdraw(msg.sender, token);
         }
         balanceStates[msg.sender][token].pendingWithdraws = PendingFlux({ amount: amount, stateIndex: getCurrentBatchId() });
         emit WithdrawRequest(msg.sender, token, amount, getCurrentBatchId());
     }
 
-    function withdraw(address token, address owner) public {
-        updateDepositsBalance(owner, token); // withdrawn amount might just be deposited before
+    function withdraw(address user, address token) public {
+        updateDepositsBalance(user, token); // withdrawn amount might just be deposited before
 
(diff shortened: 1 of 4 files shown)


josojo:
Hey Ben,

yeah, we defintiviely need something like that. But I guess gcd is too complicated. I would rather work with a norm. What do you think about 
```
|prices| = max(p_i) 0<=i<= prices.length -1
```
and 
```
require(norm(prices) = 10**18)
```
?

bh2smith:
> I would rather work with a norm.

I like the idea of a norm but why not fix the fee token at a specific value.

fleupold:
I also think having the first price always be 10**18 is easiest and should do the trick.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FOQl3QoK4Iw9FK%2Fgiphy.gif&token=keywords" data-gitmeme-token="keywords"></a>

twalth3r:
Fixing the price of the fee token to 1 seems the most easy way to go to me. And I currently don't see any reason for doing it differently. Working with a norm would require some rounding again to make it equal 10**18 exactly, I guess.

PR #254, opened 2019-10-31
249/doc strings

### LINKED ISSUE #249: [PoC] Doc Strings for MVP contracts
Should include doc strings for the following two contracts:

1. EpochTokenLocker
2. StableCoinConverter

Please use the layout suggested in the [Solidity documentation](https://solidity.readthedocs.io/en/v0.4.24/layout-of-source-files.html) which refers to the [Doxygen](https://en.wikipedia.org/wiki/Doxygen) style as in the following example:

```js
/** @title Shape calculator. */
contract shapeCalculator {
    /** @dev Calculates a rectangle's surface and perimeter.
      * @param w Width of the rectangle.
      * @param h Height of the rectangle.
      * @return s The calculated surface.
      * @return p The calculated perimeter.
      */
    function rectangle(uint w, uint h) returns (uint s, uint p) {
        s = w * h;
        p = 2 * (w + h);
    }
}
```

bh2smith:
introducing Doxygen style doc strings in the relevant PoC smart contracts. Closes #249

**Testplan:** See that unit tests are still passing.

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -6,6 +6,7 @@ import "@gnosis.pm/solidity-data-structures/contracts/libraries/IterableAppendOn
 import "solidity-bytes-utils/contracts/BytesLib.sol";
 import "./libraries/TokenConservation.sol";
 
+
 /** @title Stable Coin Converter - A decentralized exchange for stable tokens as a batch auciton.
  * @author Alex Herrmann - <alex@gnosis.pm>
  * @author Benjamin Smith - <ben@gnosis.pm>
@@ -225,7 +226,6 @@ contract StablecoinConverter is EpochTokenLocker {
         tokenConservation.checkTokenConservation();
         documentTrades(batchIndex, owners, orderIds, volumes, tokenIdsForPrice);
     }
-
     /**
      * Public View Methods
      */
@@ -282,7 +282,6 @@ contract StablecoinConverter is EpochTokenLocker {
             return 0;
         }
     }
-
     /**
      * Internal Functions
      */
@@ -415,7 +414,6 @@ contract StablecoinConverter is EpochTokenLocker {
             );
         }
     }
-
     /**
      * Private Functions
      */
@@ -430,7 +428,6 @@ contract StablecoinConverter is EpochTokenLocker {
         );
         previousSolution.objectiveValue = newObjectiveValue;
     }
-
     // Private view
 
     /** @dev Compute trade execution based on executedBuyAmount and relevant token prices
@@ -449,7 +446,6 @@ contract StablecoinConverter is EpochTokenLocker {
         );
         return (executedBuyAmount, executedSellAmount);
     }
-
     // Private pure
 
     /** @dev used to determine if an order is valid for specific auction/batch
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -187,9 +187,13 @@ contract StablecoinConverter is EpochTokenLocker {
       * 3. `tokenIdsForPrice` is sorted.
       * 4. Number of touched orders does not exceed `MAX_TOUCHED_ORDERS`.
       * 5. Each touched order is valid at current `batchIndex`.
-      * 6. `executedSellAmount` does not exceed order's remaining amount.
+      * 6. Each touched order's `executedSellAmount` does not exceed its remaining amount.
       * 7. Limit Price of each touched order is respected.
-      * 8.
+      * 8. Solution's objective evaluation must be positive.
+      *
+      * Sub Requirements: Those nested within other functions
+      * - checkAndOverrideObjectiveValue; Objetive Evaluation is greater than current winning solution
+      * - checkTokenConservation; for all, non-fee, tokens total amount sold == total amount bought
       */
     function submitSolution(
         uint32 batchIndex,
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -87,7 +87,7 @@ contract StablecoinConverter is EpochTokenLocker {
         addToken(feeToken); // fee Token will always have the token index 0
     }
 
-    /** @dev Used to list a new token on the exchange.
+    /** @dev Used to list a new token on the contract: Hence, making it available for exchange in an auction.
       * @param _tokenAddress ERC20 token to be listed.
       */
     function addToken(address _tokenAddress) public {
@@ -99,7 +99,7 @@ contract StablecoinConverter is EpochTokenLocker {
         numTokens++;
     }
 
-    /** @dev a user facing function used to place limit sell orders in auction with expiry defined by batchId
+    /** @dev A user facing function used to place limit sell orders in auction with expiry defined by batchId
       * @param buyToken id of token to be bought
       * @param sellToken id of token to be sold
       * @param validUntil - batchId represnting order's expiry
@@ -144,7 +144,8 @@ contract StablecoinConverter is EpochTokenLocker {
         emit OrderCancelation(msg.sender, id);
     }
 
-    /** @dev a user facing function used to delete expired orders
+    /** @dev A user facing function used to delete expired orders.
+      * This release of storage gives a gas refund to msg.sender and requires that all orders are expired.
       * @param ids referencing the indices of user's orders to be deleted
       */
     function freeStorageOfOrder(uint[] memory ids) public {
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -188,7 +188,7 @@ contract StablecoinConverter is EpochTokenLocker {
       *
       * Requirements:
       * - Solutions for this `batchIndex` are currently being accepted.
-      * - feeToken price is specified.
+      * - Fee Token price is non-zero
       * - `tokenIdsForPrice` is sorted.
       * - Number of touched orders does not exceed `MAX_TOUCHED_ORDERS`.
       * - Each touched order is valid at current `batchIndex`.
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -101,7 +101,6 @@ contract StablecoinConverter is EpochTokenLocker {
     /** @dev a user facing function used to place limit sell orders in auction with expiry defined by batchId
       * @param buyToken id of token to be bought
       * @param sellToken id of token to be sold
-      * @param isSellOrder - TODO - remove with
       * @param validUntil - batchId represnting order's expiry
       * @param buyAmount - relative minimum amount of requested buy amount
       * @param sellAmount - maximum amount of sell token to be exchanged
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -6,7 +6,11 @@ import "@gnosis.pm/solidity-data-structures/contracts/libraries/IterableAppendOn
 import "solidity-bytes-utils/contracts/BytesLib.sol";
 import "./libraries/TokenConservation.sol";
 
-
+/** @title Stable Coin Converter - A decentralized exchange for stable tokens as a batch auciton.
+ * @author Alex Herrmann - <alex@gnosis.pm>
+ * @author Benjamin Smith - <ben@gnosis.pm>
+ * @author Felix Leupold - <felix@gnosis.pm>
+ */
 contract StablecoinConverter is EpochTokenLocker {
     using SafeMath for uint128;
     using BytesLib for bytes32;
@@ -52,15 +56,23 @@ contract StablecoinConverter is EpochTokenLocker {
     IdToAddressBiMap.Data private registeredTokens;
 
     uint public MAX_TOKENS;  // solhint-disable var-name-mixedcase
-    uint16 public numTokens = 0;
-    uint128 public feeDenominator; // fee is (1 / feeDenominator)
+    uint16 public numTokens;
+    uint128 public feeDenominator;
 
+    /** @dev Constructor determines exchange parameters
+      * @param maxTokens The maximum number of tokens that can be listed.
+      * @param _feeDenominator fee as a proportion is (1 / feeDenominator)
+      * @param feeToken Address of ERC20 fee token.
+      */
     constructor(uint maxTokens, uint128 _feeDenominator, address feeToken) public {
         MAX_TOKENS = maxTokens;
         feeDenominator = _feeDenominator;
         addToken(feeToken); // fee Token will always have the token index 0
     }
 
+    /** @dev Used to list a new token on the exchange.
+      * @param _tokenAddress ERC20 token to be listed.
+      */
     function addToken(address _tokenAddress) public {
         require(numTokens < MAX_TOKENS, "Max tokens reached");
         require(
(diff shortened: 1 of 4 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -42,25 +46,52 @@ contract StablecoinConverter is EpochTokenLocker {
...
+        uint256 objectiveValue;
+    }
+
+    struct TradeData {
+        address owner;
+        uint128 volume;
+        uint16 orderId;
+    }
+
     // User-> Order
```
nlordell: Public members should also have docs IMO as they generate public view functions.

nlordell (changes requested):

nlordell:
Is it possible to add a solc lint for documentation of public functions (à la `#![deny(missing_docs)]` in rust)?

bh2smith:
> Is it possible to add a solc lint for documentation of public functions (à la #![deny(missing_docs)] in rust)?

Unsure, will look into it...

contracts/StablecoinConverter.sol
```diff
@@ -7,6 +7,11 @@ import "solidity-bytes-utils/contracts/BytesLib.sol";
 import "./libraries/TokenConservation.sol";
 
 
+/** @title Stable Coin Converter - A decentralized exchange for stable tokens as a batch auciton.
+ * @author Alex Herrmann - <alex@gnosis.pm>
+ * @author Benjamin Smith - <ben@gnosis.pm>
+ * @author Felix Leupold - <felix@gnosis.pm>
```
fleupold: not too keen on being personally mentioned here. Maybe we should have a dfusion team alias?

contracts/StablecoinConverter.sol
```diff
@@ -70,6 +102,15 @@ contract StablecoinConverter is EpochTokenLocker {
         numTokens++;
     }
 
+    /** @dev a user facing function used to place limit sell orders in auction with expiry defined by batchId
+      * @param buyToken id of token to be bought
+      * @param sellToken id of token to be sold
+      * @param isSellOrder - TODO - remove with
```
fleupold: Would it make sense to rebase on top of the PR that already removed this?

contracts/StablecoinConverter.sol
```diff
@@ -102,71 +143,32 @@ contract StablecoinConverter is EpochTokenLocker {
...
-                        order
-                    ));
-                }
-                if (user == allUsers.last) {
-                    stop = true;
-                } else {
-                    user = allUsers.next(user);
-                }
-            }
+    /** @dev a user facing function used to delete expired orders
```
fleupold: maybe mention that this is worth doing for the gas refund (and assumes that orders are already invalid)

contracts/StablecoinConverter.sol
```diff
@@ -224,23 +226,81 @@ contract StablecoinConverter is EpochTokenLocker {
         tokenConservation.checkTokenConservation();
         documentTrades(batchIndex, owners, orderIds, volumes, tokenIdsForPrice);
     }
+    /**
+     * Public View Methods
+     */
+
+    /** @dev View returning ID of listed tokens
+      * @param addr address of listed token.
+      * @return tokenId as stored, via BiMap, within the contract.
```
fleupold: I think the fact that we are using a BiMap is an implementation detail that we don't need to document.

contracts/StablecoinConverter.sol
```diff
@@ -224,23 +226,81 @@ contract StablecoinConverter is EpochTokenLocker {
...
+    /** @dev View returning address of listed token by ID
+      * @param id tokenId as stored, via BiMap, within the contract.
+      * @return address of (listed) token
+      */
+    function tokenIdToAddressMap(uint16 id) public view returns (address) {
+        return IdToAddressBiMap.getAddressAt(registeredTokens, id);
+    }
 
+    /** @dev View returning all currently stored, byte-encoded sell orders
+      * @return encoded bytes representing all orders ordered by (user, index)
```
fleupold, line 290: Should we elaborate on the encoding? Or is this just writing the code again?

contracts/StablecoinConverter.sol
```diff
@@ -224,23 +226,81 @@ contract StablecoinConverter is EpochTokenLocker {
...
+                    stop = true;
+                } else {
+                    user = allUsers.next(user);
+                }
+            }
+        }
+        return elements;
+    }
+
+    /** @dev called by solvers to determine their solution better than the current winner.
```
fleupold: I think there is an *if* missing in this sentence. Also not sure if we should make assumptions on why someone would call this method (there might be many other reasons).

contracts/StablecoinConverter.sol
```diff
@@ -224,23 +226,81 @@ contract StablecoinConverter is EpochTokenLocker {
...
 
+    /** @dev called at the end of submitSolution with a value of tokenConservation / 2
+      * @param feeReward amount to be rewarded to the solver
+      */
     function grantRewardToSolutionSubmitter(uint feeReward) internal {
         previousSolution.feeReward = feeReward;
         addBalanceAndBlockWithdrawForThisBatch(msg.sender, tokenIdToAddressMap(0), feeReward);
     }
 
+    /** @dev called early from within submitSolution to update the token prices.
```
fleupold: What does "early" mean? Can we be more specific. Also I'm not sure if we need to document all internal methods as well?

contracts/StablecoinConverter.sol
```diff
@@ -250,6 +310,11 @@ contract StablecoinConverter is EpochTokenLocker {
         }
     }
 
+    /** @dev used to evaluate executedSellAmount from prices and executedBuyAmount
+      * @param executedBuyAmount amount of buyToken to be obtained in an exchange
+      * @param buyTokenPrice price of buyToken
+      * @param sellTokenPrice price of sellToken
```
fleupold: Missing return type

contracts/StablecoinConverter.sol
```diff
@@ -269,14 +334,24 @@ contract StablecoinConverter is EpochTokenLocker {
         return uint128(sellAmount);
     }
 
+    /** @dev Updates an order's remaing requested sell amount upon (partial) execution of a standing order
+      * @param owner order's corresponding user address
+      * @param orderId index of order in list of owner's orders
+      * @param executedAmount proportion of order's requested sellAmount that was filled.
```
fleupold, line 356: also missing return docs

fleupold (commented):
Nice work. I'm not sure if we need to document the internal functions as well. Should we maybe also document in what cases the function reverts?

contracts/StablecoinConverter.sol
bh2smith: Yes, sure... although this author convention has been used for all other gnosis contracts.

contracts/StablecoinConverter.sol
bh2smith: I plan to merge master into this branch before continuing... yes.

contracts/StablecoinConverter.sol
bh2smith:
```suggestion
      * @return tokenId as stored within the contract.
```

contracts/StablecoinConverter.sol
bh2smith:
```suggestion
    /** @dev gets the objective value of currently winning solution.
```

contracts/StablecoinConverter.sol
bh2smith: Two options here:

1. These return strings should be concise and we wouldn't want to write the code again...
2. If we are to generate documentation based on this, then perhaps we had better see what the auto generated documentation looks like and decide if it is worth rewriting the code again.

I'd say that its best to be concise and keep these doc-strings to at most on line per @param/return.

contracts/StablecoinConverter.sol
```diff
@@ -97,71 +135,32 @@ contract StablecoinConverter is EpochTokenLocker {
...
-        uint16 orderId;
-    }
-
+    /** @dev a solver facing function called for auction settlement
+      * @param batchIndex index of auction solution is referring to
+      * @param owners array of addresses corresponding to touched orders
+      * @param orderIds array of order ids used in parallel with owners to identify touched order
+      * @param volumes executed buy amounts for each order identified by index of owner-orderId arrays
+      * @param prices prices for touched tokens indexed by next parameter
+      * @param tokenIdsForPrice ordered list of touched token ids
```
josojo: If we have the documentation here, maybe we can replace the comments in code.

btw, 
```
// fee token id not required since always 0
```
is no longer valid. Would you mind cleaning it in this PR?

contracts/StablecoinConverter.sol
```diff
@@ -40,25 +43,52 @@ contract StablecoinConverter is EpochTokenLocker {
...
+      * @param feeToken Address of ERC20 fee token.
+      */
     constructor(uint maxTokens, uint128 _feeDenominator, address feeToken) public {
         MAX_TOKENS = maxTokens;
         feeDenominator = _feeDenominator;
         addToken(feeToken); // fee Token will always have the token index 0
     }
 
+    /** @dev Used to list a new token on the exchange.
+      * @param _tokenAddress ERC20 token to be listed.
```
josojo: Maybe you could describe what listed means on our exchange

josojo (commented):
Wow, this was quick Ben. Left some small comments

contracts/StablecoinConverter.sol
bh2smith: @josojo - How does this look?

```
Used to list a new token on the contract: Hence, making it available for exchange in an auction.
```

bh2smith:
@fleupold 

> I'm not sure if we need to document the internal functions as well. 

I'll leave these in for now and experiment with [solidity-docgen](https://github.com/OpenZeppelin/solidity-docgen) to see how it looks with and without them. 

> Should we maybe also document in what cases the function reverts?

I like this idea. Do we have any suggested syntaxtic approach to recording this? Maybe something like

```js
    /** @dev
      * @param
      * @require  <----- Not sure yet how `solidity-doc` gen will treat this.
      * @return
      */
```

contracts/StablecoinConverter.sol
bh2smith: I would be happy to remove the internal method documentation. Does anyone else have any opinion here? @josojo, @nlordell?

nlordell:
This is how OpenZeppelin does it:
```
    /**
     * @dev See {IERC20-transferFrom}.
     *
     * Emits an {Approval} event indicating the updated allowance. This is not
     * required by the EIP. See the note at the beginning of {ERC20};
     *
     * Requirements:
     * - `sender` and `recipient` cannot be the zero address.
     * - `sender` must have a balance of at least `amount`.
     * - the caller must have allowance for `sender`'s tokens of at least
     * `amount`.
     */
```

contracts/StablecoinConverter.sol
nlordell: Personally, I don't think they are needed but there is also no harm in including them.

contracts/StablecoinConverter.sol
```diff
@@ -169,6 +180,16 @@ contract StablecoinConverter is EpochTokenLocker {
...
+      *
+      * Requirements:
+      * 1. Solutions for this `batchIndex` are currently being accepted.
+      * 2. feeToken price is specified.
+      * 3. `tokenIdsForPrice` is sorted.
+      * 4. Number of touched orders does not exceed `MAX_TOUCHED_ORDERS`.
+      * 5. Each touched order is valid at current `batchIndex`.
+      * 6. `executedSellAmount` does not exceed order's remaining amount.
+      * 7. Limit Price of each touched order is respected.
+      * 8.
```
fleupold: What's

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2F2pVNVd922Igak%2Fgiphy-downsized-medium.gif&token=eight" data-gitmeme-token="eight"></a>

Should we just stick to 
- bullet
- points
- ?

contracts/StablecoinConverter.sol
```diff
@@ -285,47 +341,42 @@ contract StablecoinConverter is EpochTokenLocker {
...
     function revertRemainingOrder(
         address owner,
         uint orderId,
         uint128 executedAmount
-    ) internal returns (uint) {
+    ) internal {
         orders[owner][orderId].remainingAmount = uint128(orders[owner][orderId].remainingAmount.add(executedAmount));
     }
 
+    /** @dev
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F3oEjHGrVGrqgFFknfO%2Fgiphy.gif&token=empty" data-gitmeme-token="empty"></a>

contracts/StablecoinConverter.sol
```diff
@@ -386,11 +518,21 @@ contract StablecoinConverter is EpochTokenLocker {
...
+      */
     function checkOrderValidity(Order memory order, uint batchIndex) private pure returns (bool) {
         return order.validFrom <= batchIndex && order.validUntil >= batchIndex;
     }
 
+    /** @dev Token ordering is verified by submitSolution. Required because binary search is used to fetch token info.
+      * @param tokenIdsForPrice list of tokenIds
+      * @return true if tokenIdsForPrice is sorted else false
+      */
     function checkPriceOrdering(uint16[] memory tokenIdsForPrice) private pure returns (bool) {
```
fleupold: unrelated: Can't this function go into the tokenConservation library?

fleupold (commented):
Looking pretty good already. Let me know when this is no longer a draft.

contracts/StablecoinConverter.sol
bh2smith: Ya sure... I'll fix this soon

contracts/StablecoinConverter.sol
bh2smith: Ya Probably. Should I do it after this is ready?

nlordell (approved):
Awesome!

bh2smith:
@fleupold & @josojo This is ready for review!

contracts/StablecoinConverter.sol
```diff
@@ -101,79 +154,59 @@ contract StablecoinConverter is EpochTokenLocker {
...
+      * @param batchIndex index of auction solution is referring to
+      * @param owners array of addresses corresponding to touched orders
+      * @param orderIds array of order ids used in parallel with owners to identify touched order
+      * @param volumes executed buy amounts for each order identified by index of owner-orderId arrays
+      * @param prices list of prices for touched tokens indexed by next parameter
+      * @param tokenIdsForPrice price[i] is the price for the token with tokenID tokenIdsForPrice[i]
+      *
+      * Requirements:
+      * - Solutions for this `batchIndex` are currently being accepted.
+      * - feeToken price is specified.
```
josojo: One minor and very unimportant thing in order to prove that I read it:
Upper case for feeToken

contracts/StablecoinConverter.sol
```diff
@@ -370,15 +430,89 @@ contract StablecoinConverter is EpochTokenLocker {
...
+            .sub(currentPrices[order.buyToken].mul(order.priceNumerator));
+        return leftoverSellAmount.mul(limitTerm).div(order.priceDenominator).toUint128();
+    }
+
+    /** @dev Evaluates executedBuy amount based on prices and executedBuyAmout (fees included)
+      * @param executedBuyAmount amount of buyToken executed for purchase in batch auction
+      * @param buyTokenPrice uniform clearing price of buyToken
+      * @param sellTokenPrice uniform clearing price of sellToken
+      * @return executedSellAmount as expressed in Equation (2)
+      * https://github.com/gnosis/dex-contracts/issues/173#issuecomment-526163117
```
josojo, line 477: LIke the references!

josojo (approved):
Nice Ben!

Great PR! Thanks for making the code so much cleaner.

contracts/StablecoinConverter.sol
bh2smith: What is upper case? `FeeToken`? `FEE_TOKEN`? `Fee Token`? I do not see anywhere in the contract that explicitly refers to this so I am unsure how to proceed with this one.

josojo: I would just write FeeToken instead of feeToken, as everything starts with upper case letters. Otherwise, if you want to refer to the variable name, you might wanna use 'feeToken'.

But it is really minor

bh2smith:
```suggestion
      * - Fee Token price is non-zero
```

PR #264, opened 2019-11-02
[Easy] Naming consistency (owner<->user) and parameter ordering

bh2smith:
These are changes related to the consistency of smart contract's naming conventions and parameter orderings. These changes consist of two items

1. Replace all occurrences of `owner` with `user` (since user appeared many more times than owner)

2. Switch withdraw parameters to user-token which lines up with all public view functions. 

Observe, the old `withdraw` function signature

https://github.com/gnosis/dex-contracts/blob/0f1995cc25ebd94daacdd5247bd843f194cf93ac/contracts/EpochTokenLocker.sol#L73

in comparison with all the public view functions:

https://github.com/gnosis/dex-contracts/blob/0f1995cc25ebd94daacdd5247bd843f194cf93ac/contracts/EpochTokenLocker.sol#L101-L115

Special **note** to @anxolin, as this change may affect the front end...

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -70,11 +70,11 @@ contract EpochTokenLocker {
         emit WithdrawRequest(msg.sender, token, amount, getCurrentBatchId());
     }
 
-    function withdraw(address token, address owner) public {
-        updateDepositsBalance(owner, token); // withdrawn amount might just be deposited before
+    function withdraw(address user, address token) public {
+        updateDepositsBalance(user, token); // withdrawn amount might just be deposited before
 
         require(
-            balanceStates[owner][token].pendingWithdraws.stateIndex < getCurrentBatchId(),
+            balanceStates[user][token].pendingWithdraws.stateIndex < getCurrentBatchId(),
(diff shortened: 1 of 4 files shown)


josojo (approved):
Thanks!
You see all the inconsistencies! 

Please also modify the withdraw script used in the e2e tests.

This also requires changes in the rust code. Please also make a PR for that.

bh2smith:
@josojo - Lucky for us, there is nothing to change in the `dex-services` repo, since this script uses named arguments. Observe, the only place were we call claim_withdraw in the services is here:

```
step "Claim withdraw" \
"npx truffle exec scripts/stablex/claim_withdraw.js --accountId=0 --tokenId=1 | grep \"Success! Balance of token 1 before claim: 3000000000000000000000, after claim: 3999000000000000000000\""
```

which is done by name!

However, I just noticed that these scripts are depending on the specific instance of ganache here

```
const withdrawer = accounts[argv.accountId]
```

This is not good and I will create an issue for it.

bh2smith:
> Special note to @anxolin, as this change may affect the front end...

Here.

PR #271, opened 2019-11-04
268/move checkPriceOrdering

### LINKED ISSUE #268: Move checkPriceOrdering to TokenConservation Library
The following function would be more appropriate to have in the TokenConservation libary

https://github.com/gnosis/dex-contracts/blob/0f1995cc25ebd94daacdd5247bd843f194cf93ac/contracts/StablecoinConverter.sol#L394-L401

bh2smith:
This started, at first, as a simply moving the `checkPriceOrdering` into the appropriate/relevant library where it is used but I soon realized that there was a bug involving non-uniqueness of tokens. ~This was easily fixed by making the weak inequality into a strict one and the world was at peace!~

Sorry that inequality was not changed, it was already ensuring non-uniqueness.

Closes #268

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -12,6 +12,7 @@ contract StablecoinConverter is EpochTokenLocker {
     using BytesLib for bytes32;
     using BytesLib for bytes;
     using TokenConservation for int[];
+    using TokenConservation for uint16[];
 
     uint constant private MAX_UINT128 = 2**128 - 1;
     uint constant public MAX_TOUCHED_ORDERS = 25;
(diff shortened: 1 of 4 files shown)


josojo (approved):
Ready for merge. Great. Thanks!

fleupold (approved):
> This was easily fixed by making the weak inequality into a strict one and the world was at peace!

I don't see where this happened. It seems to me that all inequalities stayed the same. Can you point out where in the PR it was changed?

bh2smith:
> I don't see where this happened. It seems to me that all inequalities stayed the same. Can you point out where in the PR it was changed?

Yes, sorry, I was wrong. The weak inequality was actually the thing ensuring uniqueness. I only introduced a unit test that reverts when sorted but not unique token IDs.

PR #272, opened 2019-11-04
Prevent Price Scaling Hack

### LINKED ISSUE #253: [PoC] Price Vector Scaling Hack Prevention
As pointed out, discussed and decided upon in #240, we must ensure `prices[0] = 10^18` (i.e. the fee token price is always normalized to "1 ETH").

bh2smith:
In reference to the accidentally closed #240 

Closes #253 

This task is going to be extremely hard and could take several days. It will involve fixing all the unit tests involving solution submission and will be a lot of tedious calculations.

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -244,6 +244,7 @@ contract StablecoinConverter is EpochTokenLocker {
     ) public {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
         require(tokenIdsForPrice[0] == 0, "fee token price has to be specified");
+        require(prices[0] == 1 ether, "fee token price must be 10^18");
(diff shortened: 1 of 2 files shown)


nlordell (approved):
Short and sweet

josojo (approved):
Now you are harvesting your hard work!

Great.

bh2smith:
Let us all note that this may not be the most optimal solution. Let me briefly explore the possibility that we assume the feeToken is always fixed at 10^18. This would imply that the solver does not need to actually pass this value. We could essentially assume that this price is fixed and could avoid the following two verification steps:

https://github.com/gnosis/dex-contracts/blob/6b14b2e74666996a64ad08a2cb492b8ff17a836d/contracts/StablecoinConverter.sol#L248-L249

Would anyone like to see this (at least as a future improvement) or not? @josojo , @fleupold, @nlordell.

PR #274, opened 2019-11-05
[Easy] allow specified batchId on withdraw requests

### LINKED ISSUE #260: [PoC] Include batchId in withdrawRequest Parameters
Such a feature would useful in order to preemptively request your withdraw for a future batch (e.g. to withdraw your sell token after validUntil in case the order doesn't get executed)

Should begin with including `batchId` into the following function parameters.

https://github.com/gnosis/dex-contracts/blob/6c4bb7dc30eb0213997dfcfbe5cdcbb0daa3b4cf/contracts/EpochTokenLocker.sol#L64-L71

**Note that** for backwards compatibility, we may want to keep the original functions as they were (so that existing scripts don't all need to be refactored).

bh2smith:
User can now specify withdraw requests for a future batch. This would/could be useful when trying to execute a sequence of `deposit, placeOrder, requestWithdraw` in a single transaction.

Note that, for backwards compatibility, the originial `requestWithdraw` function remains with its original parameters, but now (internally) makes a call to `requestFutureWithdraw` with `batchId = getCurrentBatchId()`

Closes #260 

**TestPlan** Unit tests.

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -62,12 +62,17 @@ contract EpochTokenLocker {
     }
 
     function requestWithdraw(address token, uint amount) public {
+        requestFutureWithdraw(token, amount, getCurrentBatchId());
+    }
+
+    function requestFutureWithdraw(address token, uint amount, uint batchId) public {
         // first process old pendingWithdraw, as otherwise balances might increase for currentBatchId - 1
         if (hasValidWithdrawRequest(msg.sender, token)) {
             withdraw(msg.sender, token);
         }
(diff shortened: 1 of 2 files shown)


josojo (approved):
Yes, thanks Ben.

I also convinced myself, that we need this requestFutureWithdraw in order to allow "one trade with withdraw in one transaction"

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FpCO5tKdP22RC8%2Fgiphy.gif&token=nice" data-gitmeme-token="nice"></a>

PR #275, opened 2019-11-05
241/Reduce Solution Submission Deadline (4 Minutes)

### LINKED ISSUE #241: [PoC] Solution submission only for 4 minutes
Requirement:
Solutions should be only submittable within the first 4 minutes after closing the batch. 

Motivation:
Some traders want to know how much they have traded in the last batch end only then place new orders based on this information:
Especially, if they have bought a lot of tokens, they might want to buy less of them in the new batch or sell them already in the new batch rather than waiting the next one only.

bh2smith:
These changes reflect the fact that solutions for auction batches are only accepted up to 4 minutes after the batch has closed. Note that 240 seconds is exactly 4 minutes and that this bears no relation to the constant BATCH_TIME (which happens to be 5 minutes).

⚠️ - Might have to adjust the `dex-services` to agree with this deadline.

This PR depends directly on the passing of this [https://github.com/gnosis/dex-services/pull/292](https://github.com/gnosis/dex-services/pull/292)

Closes #241

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -181,8 +181,15 @@ contract EpochTokenLocker {
     /** @dev used to determine how much time is left in a batch
       * return seconds remaining in current batch
       */
-    function getSecondsRemainingInBatch() public view returns(uint) {
-        return BATCH_TIME - (now % BATCH_TIME);
+    function getSecondsRemainingInBatch() public view returns(uint256) {
+        return BATCH_TIME - getBatchTime();
+    }
+
+    /** @dev used to determine number of seconds passed in current batch
+      * return seconds passed in current batch
+      */
+    function getBatchTime() public view returns(uint256) {
+        return now % BATCH_TIME;
     }
 
     /** @dev fetches and returns user's balance
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -182,14 +182,7 @@ contract EpochTokenLocker {
       * return seconds remaining in current batch
       */
     function getSecondsRemainingInBatch() public view returns(uint256) {
-        return BATCH_TIME - getBatchTime();
-    }
-
-    /** @dev used to determine number of seconds passed in current batch
-      * return seconds passed in current batch
-      */
-    function getBatchTime() public view returns(uint256) {
-        return now % BATCH_TIME;
+        return BATCH_TIME - (now % BATCH_TIME);
     }
 
     /** @dev fetches and returns user's balance
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -208,7 +208,7 @@ contract StablecoinConverter is EpochTokenLocker {
         uint128[] memory prices,
         uint16[] memory tokenIdsForPrice
     ) public {
-        require(batchIndex == getCurrentBatchId() - 1, "Solutions are no longer accepted for this batch");
+        require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
         require(tokenIdsForPrice[0] == 0, "fee token price has to be specified");
         require(tokenIdsForPrice.checkPriceOrdering(), "prices are not ordered by tokenId");
         require(owners.length <= MAX_TOUCHED_ORDERS, "Solution exceeds MAX_TOUCHED_ORDERS");
@@ -310,6 +310,11 @@ contract StablecoinConverter is EpochTokenLocker {
             }
         }
         return elements;
+    }
+
+    function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
+        // Note that (now % 300 < 240) corresponds to 4 minutes after the batch has closed.
+        return batchIndex == getCurrentBatchId() - 1 && now % 300 < 240;
     }
 
     /** @dev gets the objective value of currently winning solution.
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -347,7 +347,7 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
-        return batchIndex == getCurrentBatchId() - 1 && getBatchTime() < 4 minutes;
+        return batchIndex == getCurrentBatchId() - 1 && getSecondsRemainingInBatch() >= 1 minutes;
     }
 
     /** @dev gets the objective value of currently winning solution.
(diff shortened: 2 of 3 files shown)


contracts/StablecoinConverter.sol
```diff
@@ -312,6 +312,11 @@ contract StablecoinConverter is EpochTokenLocker {
         return elements;
     }
 
+    function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
+        // Note that (now % 300 < 240) corresponds to 4 minutes after the batch has closed.
+        return batchIndex == getCurrentBatchId() - 1 && now % 300 < 240;
```
fleupold: Some people would probably say this should be a constant. But I'm not sure given that it's only used once it might as well stay here.

fleupold (approved):
Can you double check that this doesn't break the services?

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F3ornkbm5HXjcIb3KYE%2Fgiphy.gif&token=fast" data-gitmeme-token="fast"></a>

contracts/StablecoinConverter.sol
nlordell: nit: from what I understand `now % 300`  is the time of the current epoch where 300 is the epoch time (or `BATCH_TIME`), shouldn't this live in `EpockTokenLocker` then? Maybe something like:
```
// StableCoinConverter.sol
    function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
         return getBatchTime(batchIndex) < 240;
    }

// EpockTokenLocker
    function getBatchTime(uint32 batchIndex) public view returns(uint256) {
        // ...
    }
```

That would also future proof it a bit where if `BATCH_TIME` ever changes then this won't have to be manually updated as well.

josojo: I think since the `StablecoinConverter` is inheriting `EpockTokenLocker`, we can use directly `BATCH_TIME`

fleupold: I agree with @nlordell that by using a `getBatchTime()` method we are not leaking the implementation detail of how batches are calculated (based on `now` rather than some fixed start time for the first batch) to the StablecoinConverter.
This way we could change the batch timing implementation under the hood without affecting the StablecoinConverter, which seems cleaner to me.

bh2smith: My only question here (regarding `getBatchTime`) is what are we expecting to be returned if the requested `batchIndex` has already passed. Perhaps `getBatchTime` should not accept any input parameter and simply return however many seconds have passed in the current batch. 

I guess this would be 

```
    function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
         return batchIndex == getCurrentBatchId() - 1 && getBatchTime() < 240;
    }
    function getBatchTime() public view returns(uint256) {
        return now % BATCH_TIME;
    }
```

bh2smith: Applied these really great suggestions! Please check it out!

contracts/StablecoinConverter.sol
```diff
@@ -347,8 +347,7 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     function acceptingSolutions(uint32 batchIndex) public view returns (bool) {
-        // Note that (now % 300 < 240) corresponds to 4 minutes after the batch has closed.
-        return batchIndex == getCurrentBatchId() - 1 && now % 300 < 240;
+        return batchIndex == getCurrentBatchId() - 1 && getBatchTime() < 4 minutes;
```
nlordell: TIL solidity has time units - super cool!

contracts/EpochTokenLocker.sol
```diff
@@ -181,8 +181,15 @@ contract EpochTokenLocker {
...
+    function getSecondsRemainingInBatch() public view returns(uint256) {
+        return BATCH_TIME - getBatchTime();
+    }
+
+    /** @dev used to determine number of seconds passed in current batch
+      * return seconds passed in current batch
+      */
+    function getBatchTime() public view returns(uint256) {
+        return now % BATCH_TIME;
     }
```
nlordell, line 186: Yup, that's about what I imagined

nlordell (approved):

contracts/StablecoinConverter.sol
fleupold: Looking at the code now, we could just reuse the `getSecondsRemainingInBatch` method (I forgot this exsited) and just make sure it's more than 60s remaining? But up to you.

fleupold (approved):
LGTM

contracts/EpochTokenLocker.sol
fleupold: As mentioned in the other comment, we could just have reused `getSecondsRemainingInBatch`

bh2smith: Yes, I can still do that if this is what we prefer.

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FZ9EvIRmLEOS3JNFeVb%2Fgiphy.gif&token=lets_do_it" data-gitmeme-token="lets_do_it"></a>

bh2smith:
Horray! Merging this because services passes (without problems) and no required contract updates! 

Here is the passing build:

https://travis-ci.org/gnosis/dex-services/builds/609156987?utm_source=github_status&utm_medium=notification

test/stablecoin_converter.js
```diff
@@ -612,18 +612,50 @@ contract("StablecoinConverter", async (accounts) => {
...
+      const batchIndex = (await stablecoinConverter.getCurrentBatchId.call()).toNumber()
+      const orderId1 = await sendTxAndGetReturnValue(stablecoinConverter.placeValidFromOrder, basicTrade.orders[0].buyToken, basicTrade.orders[0].sellToken, batchIndex + 1, batchIndex + 2, basicTrade.orders[0].buyAmount, basicTrade.orders[0].sellAmount, { from: basicTrade.orders[0].user })
+      const orderId2 = await sendTxAndGetReturnValue(stablecoinConverter.placeValidFromOrder, basicTrade.orders[1].buyToken, basicTrade.orders[1].sellToken, batchIndex + 1, batchIndex + 2, basicTrade.orders[1].buyAmount, basicTrade.orders[1].sellAmount, { from: basicTrade.orders[1].user })
+
+      const prices = basicTrade.solution.prices
+      const owner = basicTrade.solution.owners
+      const orderId = [orderId1, orderId2]
+      const volume = basicTrade.solution.volume
+      const tokenIdsForPrice = basicTrade.solution.tokenIdsForPrice
+      closeAuction(stablecoinConverter)
```
josojo, line 654: The await is missing here.

Do you want a seperate PR for this correction? Or just we just later merge my PR? https://github.com/gnosis/dex-contracts/pull/285

bh2smith: Yes, please just fix it in yours! Sorry....

PR #284, opened 2019-11-07
fix dU

bh2smith:
Addresses #276 

Will require a complete overhaul of all the solution examples before it passes.

Note that this updated calculation likes to underflow on the subtraction in `disregardedUtility`

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -499,7 +499,7 @@ contract StablecoinConverter is EpochTokenLocker {
             getBalance(user, tokenIdToAddressMap(order.sellToken))
         );
         uint256 limitTerm = currentPrices[order.sellToken].mul(order.priceDenominator)
-            .sub(currentPrices[order.buyToken].mul(order.priceNumerator));
+            .mul(feeDenominator - 1).div(feeDenominator).sub(currentPrices[order.buyToken].mul(order.priceNumerator));
         return leftoverSellAmount.mul(limitTerm).div(order.priceDenominator).toUint128();
     }
 
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -488,7 +488,7 @@ contract StablecoinConverter is EpochTokenLocker {
       * @return disregardedUtility of the order (after it has been applied)
       * Note that:
       * |disregardedUtility| = (limitTerm * leftoverSellAmount) / order.sellAmount
-      * where limitTerm = price.SellToken * order.sellAmt - order.buyAmt * price.buyToken
+      * where limitTerm = price.SellToken * order.sellAmt - order.buyAmt * price.buyToken * (1 - phi)
       * and leftoverSellAmount = order.sellAmt - execSellAmt
       * Balances and orders have all been updated so: sellAmount - execSellAmt == order.remainingAmount.
       * For correctness, we take the minimum of this with the user's token balance.
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -499,7 +499,7 @@ contract StablecoinConverter is EpochTokenLocker {
             getBalance(user, tokenIdToAddressMap(order.sellToken))
         );
         uint256 limitTerm = currentPrices[order.sellToken].mul(order.priceDenominator)
-            .mul(feeDenominator - 1).div(feeDenominator).sub(currentPrices[order.buyToken].mul(order.priceNumerator));
+            .sub(order.priceNumerator.mul(feeDenominator).div(feeDenominator-1).mul(currentPrices[order.buyToken]));
         return leftoverSellAmount.mul(limitTerm).div(order.priceDenominator).toUint128();
     }
 
--- a/test/stablecoin_converter.js
+++ b/test/stablecoin_converter.js
@@ -1378,7 +1378,7 @@ function evaluateTradeUtility(buyAmount, sellAmount, executedBuyAmount, executed
 }
 
 function disregardedUtility(buyAmount, sellAmount, executedBuyAmount, executedSellAmount, priceBuyToken, priceSellToken) {
-  const limitTerm = priceSellToken * sellAmount - priceBuyToken * buyAmount
+  const limitTerm = priceSellToken * sellAmount - priceBuyToken * feeAdded(buyAmount)
   // Note, this computation assumes bidder has sufficient balance remaining
   // Usually leftoverSellAmount = MIN(sellAmount - executedSellAmount, user.balance.sellToken)
 


contracts/StablecoinConverter.sol
```diff
@@ -499,7 +499,7 @@ contract StablecoinConverter is EpochTokenLocker {
             getBalance(user, tokenIdToAddressMap(order.sellToken))
```
nlordell, line 501: The doc comments should be updated with the new equation.

nlordell (changes requested):

contracts/StablecoinConverter.sol
```diff
@@ -499,7 +499,7 @@ contract StablecoinConverter is EpochTokenLocker {
...
         );
         uint256 limitTerm = currentPrices[order.sellToken].mul(order.priceDenominator)
-            .sub(currentPrices[order.buyToken].mul(order.priceNumerator));
+            .sub(order.priceNumerator.mul(feeDenominator).div(feeDenominator-1).mul(currentPrices[order.buyToken]));
         return leftoverSellAmount.mul(limitTerm).div(order.priceDenominator).toUint128();
```
nlordell, line 505: I noticed that we are adding a fee to the `buy_amount * buy_price` instead of removing a fee from the `sell_price * sell_amount`. I don't think this would cause any issues but I'm just mentioning it because I think that in #276 it was the other way around.

bh2smith: Yes, this is how @fleupold got around the subtraction underflows that were happening.

contracts/StablecoinConverter.sol
bh2smith: fixed.... weird place to comment though...

bh2smith:
I believe that, for now, this fix is good enough. Any additional problems with this can be fixed in the next couple of related PRs (#272 & #279).

contracts/StablecoinConverter.sol
nlordell: It didn’t let me comment on the actual line :( that’s as far up it went

nlordell (approved):
🥇

contracts/StablecoinConverter.sol
nlordell: @fleupold Actually, I just realized there might be an issue with this:
`(buy_amount * fee_denominator) / (fee_denominator - 1)` introduces some error (0.0 .. 0.998 in this case) in the division which gets multiplied by `buy_price`. This means we can get up to `(buy_price * 998) / 1000` error in that term. I think the division should move to the end.

bh2smith: >  I think the division should move to the end.

As in move it to the last operation?

josojo:
```
currentPrices[order.sellToken].mul(order.priceDenominator)
            .sub(order.priceNumerator.mul(feeDenominator).div(feeDenominator-1).mul(currentPrices[order.buyToken]))
```
Before we merge this, I would like to see some proof that this term can not become smaller than 0 (and hence .sub reverts), for orders respecting the limit price.
I think the proof would be much simpler, if we would divide later by feeDominotor in the term.

contracts/StablecoinConverter.sol
nlordell: That works for me

nlordell:
@josojo, not really a proof but here is the intuition why it should be OK:

The equation for real numbers is:
```
limitTerm = (xS * (1 - phi) * pS - xB * pB)
dU = (xS - xS_exec) * limitTerm / xS
```
And what we can control is where we place the `(1 - phi)` term in the limit term computation of the disregarded utility. Note that `(1 - phi)` is espressed in terms of the fee denominator fd and `//` here denotes integer division. Our options are:
```
(1) xS * (fd - 1) // fd * pS - xB * pB
(2) xS * pS * (fd - 1) // fd - xB * pB
(3) xS * pS - xB * fd // (fd - 1) * pB
(4) xS * pS - xB * pB * fd // (fd - 1)
```
In integer multiplication, division always introduces rounding error. In `a / b = a // b + e` the error `e` is in between `[0, (b-1)/b]` which is approximately `[0, 1)` for large enough `b`. If we want to express integer division in terms of real division with an error we can rewrite this as: `a // b = a / b - e`.

1. For (1), we would have the left term of the limit term `xS * (fd - 1) // fd * pS = ((xS * fd) / (fd - 1) - e) * pS = xS * fd / (fd - 1) * pS - e * pS` which would give us a negative error of `[0, pS)` which can be very large and cause `limitTerm` to underflow for valid solutions, so this **should not be used**.
2. For (2), we would have the left term of the limit term `xS * pS * (fd - 1) // fd = (xS * pS * fd) / (fd - 1) - e = xS * pS * fd / (fd - 1) - e` which would mean that we just truncate the nearest integer. This shouldn't cause any issues since `xB * pB` is guaranteed to be an integer so truncating `xS * pS * fd / (fd - 1)` *should not* cause valid solutions to underflow and revert. This **can be used with caveats**.
3. For (3), the solution we chose, we would have on the right term of the limit term `xB * fd / (fd - 1) * pB = ((xB * (fd - 1)) / fd - e) * pB = xB * fd / (fd - 1) * pB - e * pB` which would mean that the total limit term would be `[0, pB)` larger than expected. This means that valid solutions should not underflow and **can be used**.
4. For (4), we would have on the right term of the limit term `xB * pB * fd // (fd - 1) = (xB * pB * (fd - 1)) / fd - e = xB * pB * fd / (fd - 1) - e` which would mean that the total limit would be the closes ceiling integer to the actual expected value. This means that valid solutions should not underflow and **can be used with caveats**.

**edit**: fixed equations

josojo:
> For (2), we would have the left term of the limit term xS * pS * (fd - 1) // fd = (xS * pS * fd) / (fd - 1) - e = xS * pS * fd / (fd - 1) - e which would mean that we just truncate the nearest integer. This shouldn't cause any issues since xB * pB is guaranteed to be an integer so truncating xS * pS * fd / (fd - 1) should not cause valid solutions to underflow and revert. This can be used with caveats.

This can be an issue. Imagine xS * pS * (fd - 1) / fd returns due to inaccuracy in calculations 1234.99999999, but should actually be 1235 and then the truncation reduces it to 1234, which might be smaller than  xB * pB = 1235. I am not saying that I found such an example, but this is the risk I am seeing.

I definitively prefer solution 4, as I think it will be positive, as the rounding only happens on the substracted value, and it has the smallest rounding error.

nlordell:
@josojo 

> This can be an issue. Imagine xS * pS * (fd - 1) / fd returns due to inaccuracy in calculations 1234.99999999, but should actually be 1235 and then the truncation reduces it to 1234, which might be smaller than xB * pB = 1235.

I think this is impossible, since that would require to `xS * pS * (fd - 1) / fd == x` where `x` is an integer but `xS * pS * (fd - 1) // fd != x`. This is not possible since `xS`, `pS` and `fd` are provided as integers so `xS * pS * (fd - 1)` must also be an integer. `xS * pS * (fd - 1) / fd == x` where `x` is an integer means that `fd` divides `xS * pS * (fd - 1)` which means that `xS * pS * (fd - 1) // fd` is exact.

josojo:
I would prefer this formula:
```
xS * pS - xB * pB  // (fd - 1) * fd 
```
Because then we can easily proof.
```
xS * pS - xB * pB  // (fd - 1) * fd > 0
<=>(xB *pB // (fd - 1) * fd  // pS) * pS - xB * pB * fd // (fd - 1) >0
<=>(xB *pB  // (fd - 1) * fd // pS) * pS > xB * pB // (fd - 1) * fd 
```

Alternatively, we could also calculate the sellAmount differently in the smart contracts.

nlordell:
@josojo @bh2smith My vote is for equation 2 or 4. Although we got 3 working with the unit tests it does add quite a lot of "convenient" error which I don't really like.

josojo:
With forumla two, we have the following issue:
```
(2) xS * pS * (fd - 1) // fd - xB * pB
In case the order is matched at limit price: xS/xB *(fd - 1) / fd  =  pB/ pS
<=>(xB *pB // (fd - 1) * fd  // pS) * pS * (fd - 1) // fd - xB * pB
<=>(xB *pB  // (fd - 1) * fd // pS) * pS > xB * pB // (fd - 1) * fd 

I checked, xS is calculated in the smart contract with: (xB *pB // (fd - 1) * fd  // pS)
```
And this will not be true always.

But we could use this formula:
```
(5) 
<==> xS * pS - min((xS  * pS // fd  * (fd - 1) // pB), xB) * pB // (fd - 1) * fd 
```

nlordell:
~~@josojo Sorry for the confusing notation, but here `xS` is order sell amount and `xS_exec` is executed sell amount. In the `limitTerm` for the disregarded utility, only order amounts and not executed amounts are used.~~
**edit**: yeah that could be a problem...

josojo:
I tried to derive my formula more cleanly, and I found that there is an easier and better solution:
https://github.com/gnosis/dex-contracts/issues/276#issuecomment-552741147

In this comment, I am suggesting to go with:
```python
uint256 limitTermLeft = currentPrices[order.sellToken].mul(order.priceDenominator)
uint256 limitTermRight = order.priceNumerator.mul(feeDenominator).div(feeDenominator-1).mul(currentPrices[order.buyToken])
if( limitTermLeft > limitTermRight) {
       limitTerm = limitTermLeft.sub(limitTermRight)
} else {
    limitTerm = 0
}
```

nlordell:
I like this solution - it means that disregarded utility can't be negative. This should be fine since we check that the solution is valid, limit prices respected, etc. elsewhere and if we get to the disregarded utility calculations then all should be valid.

I have one question about the order of operations, should the division be at the end instead of the middle so we incur less error? Something like:
```
uint256 limitTermLeft = currentPrices[order.sellToken].mul(order.priceDenominator)
uint256 limitTermRight = order.priceNumerator.mul(currentPrices[order.buyToken]).mul(feeDenominator).div(feeDenominator-1)
if( limitTermLeft > limitTermRight) {
       limitTerm = limitTermLeft.sub(limitTermRight)
} else {
    limitTerm = 0
}
```

The error introduced by making `.mul(currentPrices[order.buyToken])` the last operation would mean that `limitTerm` will be `[0, currentPrices[order.buyToken])` greater than the actual value - making the disregarded utility larger than it should be.

josojo:
yeah @nlordell 
I am in favor of your suggestion!

PR #295, opened 2019-11-13
[Easy] Renaming in package.json

josojo:
@anxolin suggested to do this renaming, as we have used this style in all our other projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "dex-contracts",
+  "name": "@gnosis.pm/dex-contracts",
   "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
ok!

nlordell (approved):

anxolin:
Cool, I think with this, we can do a `npm publish` already!

```bash
npm version mintor/major/patch
npm publish --access=public
```

what do you thing?

Only thing to make sure is .npmignore does not exclude the `build` dir, so we have the compiled contracts. Also, to create a `preversion` script in the `package.json`, for example:
image

What do you think?

PR #302, opened 2019-11-15
Add the StableX e2e Test Example

nlordell:
Added the stableX e2e test example to the generated auction examples (mostly for documentation and to validate the naive solver solution). This also adds some debugging facilities to the `generateTestCase` method that prints additional information when the solution is deemed invalid.

### Test plan
Didn't break any of the unit tests :crossed_fingers:

### DIFF
--- a/test/resources/math.js
+++ b/test/resources/math.js
@@ -148,12 +148,13 @@ function orderDisregardedUtility(order, executedBuyAmount, prices) {
  * @return {BN} The solution's objective value
  */
 function solutionObjectiveValue(orders, solution) {
-  return solutionObjectiveValueComputation(orders, solution).result
+  return solutionObjectiveValueComputation(orders, solution, true).result
 }
 
 /**
  * @typedef ObjectiveValueComputation
  * @type {object}
+ * @property {BN[][]} orderTokenConservation The token conservation per token per order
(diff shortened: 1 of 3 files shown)


bh2smith (approved):
Sweet!

josojo (approved):
I like it!

test/resources/math.js
```diff
@@ -148,12 +148,13 @@ function orderDisregardedUtility(order, executedBuyAmount, prices) {
  * @return {BN} The solution's objective value
  */
 function solutionObjectiveValue(orders, solution) {
-  return solutionObjectiveValueComputation(orders, solution).result
+  return solutionObjectiveValueComputation(orders, solution, true).result
```
bh2smith, line 151: Do we want this to be kept here? or remove the true?
```suggestion
  return solutionObjectiveValueComputation(orders, solution).result
```

bh2smith (approved):
Look great. Just one suggestion.

test/resources/math.js
nlordell: The true was on purpose, the `solutionObjectiveValue` function wants to be strict. I made it explicit so that if the default parameter value changes, `solutionObjectiveValue` will keep the same behaviour.

PR #310, opened 2019-11-20
Burn half the fees after auction settles

bh2smith:
As described in the specification, the auction should burn half of the fee token conservation while rewarding the solver with the other half. This PR implements exactly that. Particularly, the fees are recorded in the solution data, and upon the submission of the next solution (only in a future batch), if the burnable fees from the previous auction have not yet been burnt, they are.

**Test Plan:** See the new unit test which goes through the entire process. That is, 

- One auction passes with a non-trivial solution (`basicTrade`)
- Orders are placed for a later auction and another, valid, non-trivial solution is submitted. 
- burntFees from previous auction are indeed burnt.

### DIFF
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -391,6 +391,8 @@ contract StablecoinConverter is EpochTokenLocker {
       * @param feeReward amount to be rewarded to the solver
       */
     function grantRewardToSolutionSubmitter(uint256 feeReward) internal {
+        // Burn the difference from previous solution
+        feeToken.burnOWL(address(this), feeReward - latestSolution.feeReward);
         latestSolution.feeReward = feeReward;
         addBalanceAndBlockWithdrawForThisBatch(msg.sender, tokenIdToAddressMap(0), feeReward);
     }
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -238,12 +238,12 @@ contract StablecoinConverter is EpochTokenLocker {
         uint16[] memory tokenIdsForPrice
     ) public returns (uint256) {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
-        burnPreviousAuctionFees();
         require(claimedObjectiveValue > getCurrentObjectiveValue(), "Claimed objective is not more than current solution");
         require(tokenIdsForPrice[0] == 0, "fee token price has to be specified");
         require(prices[0] == 1 ether, "fee token price must be 10^18");
         require(tokenIdsForPrice.checkPriceOrdering(), "prices are not ordered by tokenId");
         require(owners.length <= MAX_TOUCHED_ORDERS, "Solution exceeds MAX_TOUCHED_ORDERS");
+        burnPreviousAuctionFees();
         undoCurrentSolution();
         updateCurrentPrices(prices, tokenIdsForPrice);
         delete latestSolution.trades;
--- a/contracts/StablecoinConverter.sol
+++ b/contracts/StablecoinConverter.sol
@@ -101,6 +101,7 @@ contract StablecoinConverter is EpochTokenLocker {
     constructor(uint256 maxTokens, uint128 _feeDenominator, address _feeToken) public {
         MAX_TOKENS = maxTokens;
         feeToken = TokenOWL(_feeToken);
+        feeToken.approve(address(this), uint(-1));
         feeDenominator = _feeDenominator;
         addToken(_feeToken); // feeToken will always have the token index 0
     }
@@ -237,12 +238,13 @@ contract StablecoinConverter is EpochTokenLocker {
         uint16[] memory tokenIdsForPrice
     ) public returns (uint256) {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
+        burnPreviousAuctionFees();
         require(claimedObjectiveValue > getCurrentObjectiveValue(), "Claimed objective is not more than current solution");
         require(tokenIdsForPrice[0] == 0, "fee token price has to be specified");
         require(prices[0] == 1 ether, "fee token price must be 10^18");
         require(tokenIdsForPrice.checkPriceOrdering(), "prices are not ordered by tokenId");
         require(owners.length <= MAX_TOUCHED_ORDERS, "Solution exceeds MAX_TOUCHED_ORDERS");
-        undoCurrentSolution(batchIndex);
+        undoCurrentSolution();
         updateCurrentPrices(prices, tokenIdsForPrice);
         delete latestSolution.trades;
         int[] memory tokenConservation = new int[](prices.length);
@@ -386,15 +388,21 @@ contract StablecoinConverter is EpochTokenLocker {
         allUsers.insert(msg.sender);
         return orders[msg.sender].length - 1;
     }
+
     /** @dev called at the end of submitSolution with a value of tokenConservation / 2
       * @param feeReward amount to be rewarded to the solver
       */
     function grantRewardToSolutionSubmitter(uint256 feeReward) internal {
-        // Burn the difference from previous solution and this.
-        feeToken.approve(address(this), feeReward - latestSolution.feeReward);
-        feeToken.burnOWL(address(this), feeReward - latestSolution.feeReward);
         latestSolution.feeReward = feeReward;
         addBalanceAndBlockWithdrawForThisBatch(msg.sender, tokenIdToAddressMap(0), feeReward);
+    }
+
+    /** @dev called during solution submission to burn fees from previous auction
+      */
+    function burnPreviousAuctionFees() internal {
+        if (!currentBatchHasSolution()) {
+            feeToken.burnOWL(address(this), latestSolution.feeReward);
+        }
     }
 
     /** @dev Called from within submitSolution to update the token prices.
@@ -466,10 +474,9 @@ contract StablecoinConverter is EpochTokenLocker {
     }
 
     /** @dev reverts all relevant contract storage relating to an overwritten auction solution.
-      * @param batchIndex index of referenced auction
       */
-    function undoCurrentSolution(uint32 batchIndex) internal {
-        if (latestSolution.batchId == batchIndex) {
+    function undoCurrentSolution() internal {
+        if (currentBatchHasSolution()) {
             for (uint256 i = 0; i < latestSolution.trades.length; i++) {
                 address owner = latestSolution.trades[i].owner;
                 uint256 orderId = latestSolution.trades[i].orderId;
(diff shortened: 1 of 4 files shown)


test/stablex/stablecoin_converter.js
```diff
@@ -46,6 +46,7 @@ contract("StablecoinConverter", async (accounts) => {
   let BATCH_TIME
   before(async () => {
     const feeToken = await MockContract.new()
+    await feeToken.givenAnyReturnBool(true)
```
fleupold, line 49: Sorry about that should be addressed by https://github.com/gnosis/mock-contract/pull/17

test/stablex/stablecoin_converter.js
```diff
@@ -1253,6 +1307,11 @@ contract("StablecoinConverter", async (accounts) => {
     })
   })
   describe("getEncodedAuctionElements()", async () => {
+    it("returns null when no orders.", async () => {
```
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FSIXreftELj2N2%2Fgiphy.gif&token=unrelated" data-gitmeme-token="unrelated"></a>

test/stablex/stablecoin_converter.js
```diff
@@ -467,6 +466,62 @@ contract("StablecoinConverter", async (accounts) => {
       assert.equal((await stablecoinConverter.getBalance.call(user_2, erc20_2)).toString(), basicTrade.deposits[1].amount.sub(getExecutedSellAmount(fullBuyVolumes[1], prices[0], prices[1])).toString(), "Sold tokens were not adjusted correctly")
       assert.equal((await stablecoinConverter.getBalance.call(user_2, feeToken)), fullBuyVolumes[1].toString(), "Bought tokens were not adjusted correctly")
     })
+    it("Ensure fee token is burnt as expected", async () => {
+      // Fee token can't be a Mock here.
```
fleupold: Why not? It should be able to be mocked using `givenAnyReturn(true)`. And later we can inspect how it was called using https://github.com/gnosis/mock-contract/#inspect-invocations

test/stablex/stablecoin_converter.js
```diff
@@ -467,6 +466,62 @@ contract("StablecoinConverter", async (accounts) => {
       assert.equal((await stablecoinConverter.getBalance.call(user_2, erc20_2)).toString(), basicTrade.deposits[1].amount.sub(getExecutedSellAmount(fullBuyVolumes[1], prices[0], prices[1])).toString(), "Sold tokens were not adjusted correctly")
       assert.equal((await stablecoinConverter.getBalance.call(user_2, feeToken)), fullBuyVolumes[1].toString(), "Bought tokens were not adjusted correctly")
     })
+    it("Ensure fee token is burnt as expected", async () => {
```
fleupold: nit: as expected is not very descriptive as the reader might not now what is expected. Should we say "Ensure half of the fee is burnt"?

test/stablex/stablecoin_converter.js
```diff
@@ -467,6 +466,62 @@ contract("StablecoinConverter", async (accounts) => {
...
+      await makeDeposits(stablecoinConverter, accounts, tradeExample.deposits)
+      const batchIndex = (await stablecoinConverter.getCurrentBatchId.call()).toNumber()
+      const orderIds = await placeOrders(stablecoinConverter, accounts, basicTrade.orders, batchIndex + 1)
+      await closeAuction(stablecoinConverter)
+
+      const solution = solutionSubmissionParams(basicTrade.solutions[0], accounts, orderIds)
+      await stablecoinConverter.submitSolution(batchIndex, solution.objectiveValue, solution.owners, solution.touchedOrderIds, solution.volumes, solution.prices, solution.tokenIdsForPrice, { from: solver })
+      await closeAuction(stablecoinConverter)
+
+      // Second Auction
```
fleupold, line 506: Should we maybe also check that applying a better solution doesn't burn the fees again?

fleupold (commented):
Looking good! I think the tests could be a bit more concise and testing the scenario where we have two solutions in one batch.

contracts/StablecoinConverter.sol
```diff
@@ -395,6 +397,14 @@ contract StablecoinConverter is EpochTokenLocker {
         addBalanceAndBlockWithdrawForThisBatch(msg.sender, tokenIdToAddressMap(0), feeReward);
     }
 
+    /** @dev called during solution submission to burn fees from previous auction
+      */
+    function burnPreviousAuctionFees() internal {
+        if (!currentBatchHasSolution()) {
+            feeToken.burnOWL(address(this), latestSolution.feeReward);
```
nlordell, line 404: Interesting, you have to approve to burn your own OWL?

test/stablex/stablecoin_converter.js
bh2smith: I was under the impression that I would need to verify the balances of this token and would, thus, need to deploy an actual token that could actually return valid information.

contracts/StablecoinConverter.sol
```diff
@@ -237,12 +238,13 @@ contract StablecoinConverter is EpochTokenLocker {
         uint16[] memory tokenIdsForPrice
     ) public returns (uint256) {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
+        burnPreviousAuctionFees();
```
nlordell: I'm just curious, is there a reason for this being placed here? It feels kind of unnatural to me, but then again I am a noob at smart contract coding and want to learn :smiley:.

test/stablex/stablecoin_converter.js
bh2smith:
```suggestion
    it("ensures half of the token imbalance (fees) is burned", async () => {
```
How about this @fleupold?

test/stablex/stablecoin_converter.js
bh2smith: Yes, perhaps. Will do.... Done.

nlordell (approved):
Looks good.

contracts/StablecoinConverter.sol
bh2smith: My train of thought when putting it here was that it we should at least verify that the solution being submitted it relevant for the current batch. Now that I think about it, we would probably also want  to check the claimed objective value first. Technically, whatever changes are made during this transaction will revert if any of the requirements are not satisfied. I am now considering moving this to the end where all of the state information is actually updated. Nice! I hope this helps with the learning.

contracts/StablecoinConverter.sol
bh2smith: Super weird...

contracts/StablecoinConverter.sol
josojo: I would vote for pushing it down by one line
Reasoning: first make all the solution submission gas saving tests: acceptingSolutions && claimObjectiveValue> currentObjective value, and then we start with

bh2smith: > and then we start with...

The suspense is killing me!

test/stablex/stablecoin_converter.js
bh2smith: I will remove this change. 🐼

test/stablex/stablecoin_converter.js
josojo: I think, it would be better to split this test case into two separate ones. One where you are testing that fees are burnt, if we hit a new auction, and one where you are testing that no fee's are burnt, if we are still in the same auction

bh2smith: Yes, I would agree that there should be two tests, but I don't this test is currently doing both. I will add another.

test/stablex/stablecoin_converter.js
bh2smith: Very nice! I will wait till its merged and try it out.

contracts/StablecoinConverter.sol
bh2smith: I have moved just below the first group of `requirements` (for readability).

test/stablex/stablecoin_converter.js
bh2smith: > split this test case into two separate ones 

After splitting this into two test cases, I noticed a substantial amount of duplicated code. It would literally be a copy-paste of the first test (50 lines) followed by this little snippet:

```js
      // Better second solution
      const betterSolution = solutionSubmissionParams(secondTradeExample.solutions[1], accounts, secondOrderIds)
      // This is where the first auction's fees should be burned!
      await stablecoinConverter.submitSolution(
        nextBatchIndex,
        betterSolution.objectiveValue,
        betterSolution.owners,
        betterSolution.touchedOrderIds,
        betterSolution.volumes,
        betterSolution.prices,
        betterSolution.tokenIdsForPrice,
        { from: solver })
      const afterBetterSolutionFeeBalance = await owlProxy.balanceOf(stablecoinConverter.address)
      assert(initialFeeTokenBalance.sub(basicTrade.solutions[0].burntFees).eq(afterBetterSolutionFeeBalance))
```

I would argue that there is no need for a second test just for this last little bit if the first test verifies what we want.

test/stablex/stablecoin_converter.js
bh2smith: As a separate PR.

test/stablex/stablecoin_converter.js
```diff
@@ -467,6 +466,77 @@ contract("StablecoinConverter", async (accounts) => {
       assert.equal((await stablecoinConverter.getBalance.call(user_2, erc20_2)).toString(), basicTrade.deposits[1].amount.sub(getExecutedSellAmount(fullBuyVolumes[1], prices[0], prices[1])).toString(), "Sold tokens were not adjusted correctly")
       assert.equal((await stablecoinConverter.getBalance.call(user_2, feeToken)), fullBuyVolumes[1].toString(), "Bought tokens were not adjusted correctly")
     })
+    it("ensures half of the token imbalance (fees) is burned and that better solutions don't double-burn", async () => {
```
fleupold, line 469: I agree with @josojo  that the "and" in the test description is an indicator that it should maybe be 2 tests. Also to make it more exhaustive we could submit two solutions even in the first auction (to make sure we are burning the larger amoun). But no strong feelings.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FNHGZOqNepShCE%2Fgiphy.gif&token=awesome_sauce" data-gitmeme-token="awesome_sauce"></a>

test/stablex/stablecoin_converter.js
bh2smith: Yes! I like that idea, although please see the following response regarding code duplication.

https://github.com/gnosis/dex-contracts/pull/310#discussion_r349038603

bh2smith: I will create a separate issue in correspondence with this "Nice to Have".

josojo:
Thanks for tackling this issue cleanly!

PR #315, opened 2019-11-22
[Easy] Update mock contract and simplify usage accordingly

bh2smith:
According to the most recent changes to mock contract our mock contracts should no longer require the following `await mock.givenAnyReturnBool(true)`.


bh2smith:
Closing because `true`isn't a default value for boolean and we should not expect as much.

PR #323, opened 2019-11-26
Prepare first NPM package

anxolin:
Hi! Nick and I were reviewing how to share JS pieces among the projects. And the current proposal we are going to work on is:
* `dex-contract`: ABI, TS definitions, and even utils close to the contract, like unpack orders
* `dex-js`: utilities that are used in Front ends, bots, services, adds some goodies like formatting, etc. Also re-export the utils from dex-contract so dex-js is the only thing you need to integrate using JS


package.json
```diff
@@ -38,7 +41,8 @@
     "url": "https://github.com/gnosis/dex-contracts/issues"
   },
   "homepage": "https://github.com/gnosis/dex-contracts#readme",
-  "dependencies": {
+  "dependencies": {},
```
fleupold: Can we update the README and CI scripts to indicate that we now need to do npm install dev dependencies?

bh2smith:
I find it extremely strange that the coverage has decreased here. I think some of the tests must have been skipped because certain dependencies were not available at the time...

bh2smith:
Actually it appears nothing ran!

Screenshot 2019-11-26 at 3 50 33 PM

package.json
nlordell: AFAIK `npm install` installs all dependencies including dev dependencies. As for CI, Travis runs `npm ci` for installing dependencies which also installs all dev dependencies. From what I understand vev dependencies only do not get installed when `dex-contract` is itself used as a dependency as an NPM package (which assumes the package was already built and the dev dependencies are no longer needed).

nlordell:
I think the problem is the added `yarn.lock` which confuses travis.

anxolin:
> I think the problem is the added yarn.lock which confuses travis.

I'll delete it, it shouldn't be there in this PR. Anyways, what do you guys think about using yarn instead of NPM? 
We really don't need to change that much (yarn.lock instead of the npm one), it would still be compatible with NPM, so trivial to go back anytime, and it has some small advantages to use yarn. Besides, we did the change for all the other projects :)

nlordell:
@anxolin I created an issue for it  #326

bh2smith (approved):
This is looking really great! My only question now is "will this change the way that us developers interact with the project?"

LICENSE
```diff
@@ -0,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
```
fleupold: "Adding a licence" should be a separate PR

package.json
```diff
@@ -33,45 +36,44 @@
...
   ],
   "author": "GNOSIS",
-  "license": "ISC",
+  "license": "LGPL-3.0",
   "bugs": {
     "url": "https://github.com/gnosis/dex-contracts/issues"
   },
   "homepage": "https://github.com/gnosis/dex-contracts#readme",
-  "dependencies": {
+  "dependencies": {},
```
fleupold: nit. Can this be remove then?

fleupold (approved):
Please consider using multiple PRs for orthogonal changes (we'd likely want to keep the Licence, even if we reverted the NPM package refactoring)

fleupold:
> My only question now is "will this change the way that us developers interact with the project?"

As far as I understand, the answer is No, since npm ci/install already install all devDependencies by default. It will just make projects depending on dex-contracts install less dependencies.

anxolin:
> Please consider using multiple PRs for orthogonal changes (we'd likely want to keep the Licence, even if we reverted the NPM package refactoring)

I can remove the file from this PR, but is not orthogonal. It was required to have a license to publish in NPM. I understand and agree is better two do 2 PRs, and if it's important I'll do this, no problem at all, just since it's so small and I need to do two chained PRS I thought it was ok to include it.

package.json
anxolin: sure

fleupold:
Of course it is not very important to split these up in this particular case. But I feel we are having multi-intentional PRs too frequently and we should make it a habit to not "quickly add a thing here and there" because using git is too painful (I'm sure there are plenty of ways to boost our git efficiency).

It helps finding which PRs could have caused a specific regression or problem by looking at the titles and allows reverting features individually.

LICENSE
nlordell: #334

nlordell:
Suggestions have been addressed :rocket:

josojo (approved):
Great. Thanks, that you guys figured a good way to do this!

PR #338, opened 2019-11-28
[Easy] Use `closeAuction` In `EpochTokenLocker` Tests

nlordell:
Recently master was briefly red because of a failed `EpochTokenLocker` test. The test is failrly straightforward and does not involve much code and it looks like it was failing because of unlucky (or lucky, depending on how you look at it :smiley:) timing where the new balance was being added right before the end of an epoch, which caused the following withdraw to succeed since it was in a new epoch.

This is an attempt to fix this issue by using `closeAcution` to fast-forward to the beginning of the next epoch leading to more stable unit tests.

Link to broken build and failing test: https://travis-ci.org/gnosis/dex-contracts/builds/617735468#L1596

### Test Plan

CI

### DIFF
--- a/test/stablex/epoch_token_locker.js
+++ b/test/stablex/epoch_token_locker.js
@@ -4,17 +4,10 @@ const MockContract = artifacts.require("MockContract")
 const ERC20Interface = artifacts.require("ERC20")
 
 const truffleAssert = require("truffle-assertions")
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Very good catch!

josojo (approved):

PR #340, opened 2019-11-28
Prettier JS

bh2smith:
This is a first pass over prettier JS. 
Not sure yet about these changes.. Seems to ignore some of our standards so I believe we should pull a few more rules from our `.eslintrc` into our cool new `.prettierrc`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "solhint": "solhint \"contracts/**/*.sol\"",
     "prettier:solidity": "prettier --write 'contracts/**/*.sol'",
     "prettier:js": "prettier --write './**/*.js'",
-    "pretty-check": "prettier --check 'contracts/**/*.sol'"
+    "pretty-check": "prettier --check 'contracts/**/*.sol' && prettier --check './**/*.js'"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "verify-stablex": "truffle run verify StablecoinConverter",
     "lint": "eslint .",
     "solhint": "solhint \"contracts/**/*.sol\"",
-    "prettier": "prettier --write 'contracts/**/*.sol'",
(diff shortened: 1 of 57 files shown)


josojo (approved):
Very cool, I like it a lot.

Especially with the auto-formating while saving js files, it is very convenient!

package.json
```diff
@@ -17,7 +17,8 @@
     "verify-stablex": "truffle run verify StablecoinConverter",
     "lint": "eslint .",
     "solhint": "solhint \"contracts/**/*.sol\"",
-    "prettier": "prettier --write 'contracts/**/*.sol'",
+    "prettier:solidity": "prettier --write 'contracts/**/*.sol'",
+    "prettier:js": "prettier --write './**/*.js'",
     "pretty-check": "prettier --check 'contracts/**/*.sol'"
```
nlordell: We probably want to `prettier check` JS as well.

nlordell (approved):
Love it when the computer does my work for me! :+1: :+1:. By the time you’re done with this repo we won’t need developers anymore and our tools will code for us.

test/stablex/stablecoin_converter.js
```diff
@@ -252,10 +247,7 @@ contract("StablecoinConverter", async (accounts) => {
       const currentStateIndex = await stablecoinConverter.getCurrentBatchId()
 
       const id = await sendTxAndGetReturnValue(stablecoinConverter.placeOrder, 0, 1, currentStateIndex + 3, 10, 20)
-      await truffleAssert.reverts(
-        stablecoinConverter.freeStorageOfOrders([id]),
-        "Order is still valid"
-      )
+      await truffleAssert.reverts(stablecoinConverter.freeStorageOfOrder([id]), "Order is still valid")
```
bh2smith:
```suggestion
      await truffleAssert.reverts(stablecoinConverter.freeStorageOfOrders([id]), "Order is still valid")
```

PR #344, opened 2019-11-30
Refactor Shared Computation of Minimum Required Objective Value Improvement

bh2smith:
As suggested and discussed in https://github.com/gnosis/dex-contracts/pull/342#discussion_r352186398, there was some redundancy introduced by a not so helpful helper function. This included double check of objective inequalities with extremely similar require statements.

To eliminate this, these changes compare the claimed value against the current at the beginning (for gas savings) and then verify that the claimed value equals the internally computed one. Our unit tests required only small adjustments here:

- One was removed because it was redundant
- One was adjusted/renamed to make up for the removed one.
- One had to be completely reworked, because this change uncovered an evaluation bug. That is, our externally computed objective value did not agree with the internal one. The disagreement here was because the JS evaluation of `disgregardedUtility` does not compensate for the account balances. Appropriate TODOs have been included in places where we could improve, but this is better reserved as the subject for a different PR.

### DIFF
--- a/.solhint.json
+++ b/.solhint.json
@@ -4,7 +4,7 @@
     "indent": ["error", 4],
     "quotes": ["error", "double"],
     "max-line-length": ["error", 129],
-    "function-max-lines": ["error", 62],
+    "function-max-lines": ["error", 60],
     "max-states-count": ["error", 18],
     "compiler-fixed": true,
     "no-simple-event-func-name": true,
--- a/.solhint.json
+++ b/.solhint.json
@@ -4,7 +4,7 @@
     "indent": ["error", 4],
     "quotes": ["error", "double"],
     "max-line-length": ["error", 129],
-    "function-max-lines": ["error", 60],
+    "function-max-lines": ["error", 62],
     "max-states-count": ["error", 18],
     "compiler-fixed": true,
     "no-simple-event-func-name": true,
--- a/contracts/BatchExchange.sol
+++ b/contracts/BatchExchange.sol
@@ -216,6 +216,7 @@ contract BatchExchange is EpochTokenLocker {
       *
       * Requirements:
       * - Solutions for this `batchIndex` are currently being accepted.
+      * - Claimed objetive value is a great enough improvement on the current winning solution
       * - Fee Token price is non-zero
       * - `tokenIdsForPrice` is sorted.
       * - Number of touched orders does not exceed `MAX_TOUCHED_ORDERS`.
@@ -225,7 +226,7 @@ contract BatchExchange is EpochTokenLocker {
       * - Solution's objective evaluation must be positive.
       *
       * Sub Requirements: Those nested within other functions
-      * - checkAndOverrideObjectiveValue; Objetive Evaluation is greater than current winning solution
+      * - checkAndOverrideObjectiveValue; Objetive value is a great enough improvement on the current winning solution
       * - checkTokenConservation; for all, non-fee, tokens total amount sold == total amount bought
       */
     function submitSolution(
@@ -239,7 +240,7 @@ contract BatchExchange is EpochTokenLocker {
     ) public returns (uint256) {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
         require(
-            claimedObjectiveValue.mul(IMPROVEMENT_DENOMINATOR) > getCurrentObjectiveValue().mul(IMPROVEMENT_DENOMINATOR + 1),
+            checkProposedObjectiveValue(claimedObjectiveValue),
             "Claimed objective doesn't sufficiently improve current solution"
         );
         require(verifyAmountThreshold(prices), "At least one price lower than AMOUNT_MINIMUM");
@@ -292,9 +293,7 @@ contract BatchExchange is EpochTokenLocker {
         // burntFees ensures direct trades (when available) yield better solutions than longer rings
         uint256 objectiveValue = utility.add(burntFees).sub(disregardedUtility);
         tokenConservation.checkTokenConservation();
-        // Verify and override objective value
-        require(objectiveValue == claimedObjectiveValue, "Computed objective must agree with claimed");
-        latestSolution.objectiveValue = objectiveValue;
+        checkAndOverrideObjectiveValue(objectiveValue);
         grantRewardToSolutionSubmitter(burntFees);
         documentTrades(batchIndex, owners, orderIds, buyVolumes, tokenIdsForPrice);
         return (objectiveValue);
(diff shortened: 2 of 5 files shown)


nlordell:
Just a heads up, this means we will have no way to query the objective value from the solution which is used by the dex services.

bh2smith:
> Just a heads up, this means we will have no way to query the objective value from the solution which is used by the dex-services.

Damn it! Another good point! Now what should we do?

.solhint.json
```diff
@@ -4,7 +4,7 @@
     "indent": ["error", 4],
     "quotes": ["error", "double"],
     "max-line-length": ["error", 129],
-    "function-max-lines": ["error", 60],
+    "function-max-lines": ["error", 62],
```
fleupold: Are we going to keep adjusting this in every PR? Maybe we should either disable it or chose a reasonably large value?

fleupold (commented):
> Now what should we do?

Maybe just putting the computation into a helper function (`isSignificantlyBetter`) or something alike is good enough for the moment.

.solhint.json
nlordell: Or refactor :rocket:

bh2smith: Jokes. Yes I would be happy to carefully evaluate the existing functionality of the submit solution and potentially break it into something smaller and more reasonable. Maybe even a modifier for all the initial/elementary constraints

nlordell:
commandeering this, Is that OK with you guys? @fleupold @josojo

fleupold:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2Fl3fQtuI4qruW1JERy%2Fgiphy.gif&token=commandeer" data-gitmeme-token="commandeer"></a>

.solhint.json
fleupold: @nlordell any chance we can get rid of this change?

contracts/BatchExchange.sol
```diff
@@ -577,13 +578,10 @@ contract BatchExchange is EpochTokenLocker {
...
     /** @dev determines if value is better than currently and updates if it is.
-      * @param newObjectiveValue proposed value to be updated if greater than current.
+      * @param newObjectiveValue proposed value to be updated if a great enough improvement on the current objective value
       */
     function checkAndOverrideObjectiveValue(uint256 newObjectiveValue) private {
-        require(
-            newObjectiveValue.mul(IMPROVEMENT_DENOMINATOR) > getCurrentObjectiveValue().mul(IMPROVEMENT_DENOMINATOR + 1),
-            "New objective doesn't sufficiently improve current solution"
-        );
+        require(checkProposedObjectiveValue(newObjectiveValue), "New objective doesn't sufficiently improve current solution");
```
fleupold: naming nit: here we are not checking the proposed objective value but the calculated one.

I'd also slightly prefer a method that starts with `is` (e.g. `isGoodEnoughSolution`) or the like as it conveys in the name that the function returns a bool (whereas check could also mean it just asserts it).

contracts/BatchExchange.sol
```diff
@@ -291,9 +292,9 @@ contract BatchExchange is EpochTokenLocker {
         require(utility.add(burntFees) > disregardedUtility, "Solution must be better than trivial");
         // burntFees ensures direct trades (when available) yield better solutions than longer rings
         uint256 objectiveValue = utility.add(burntFees).sub(disregardedUtility);
+        tokenConservation.checkTokenConservation();
```
fleupold: conceptually I feel this change should be all the way up just after the tokenConservation has been calculated. I understand that this is not possible as some of our test cases wouldn't assert in the right spot, but I think this is a smell with our test suite.

contracts/BatchExchange.sol
```diff
@@ -291,9 +292,9 @@ contract BatchExchange is EpochTokenLocker {
         require(utility.add(burntFees) > disregardedUtility, "Solution must be better than trivial");
```
fleupold: Is this still needed?  checkProposedObjectiveValue makes sure we are strictly better than the previous solution, right?

contracts/BatchExchange.sol
nlordell: Changed it.

contracts/BatchExchange.sol
nlordell: I reverted back to where it originally was.

nlordell: I moved this back to where it was.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3oz8xHTZ99bSDV7euc%2Fgiphy.gif&token=so_clean" data-gitmeme-token="so_clean"></a>

contracts/BatchExchange.sol
```diff
@@ -216,6 +216,7 @@ contract BatchExchange is EpochTokenLocker {
       *
       * Requirements:
       * - Solutions for this `batchIndex` are currently being accepted.
+      * - Claimed objetive value is a great enough improvement on the current winning solution
```
bh2smith, line 219: Great catch!

josojo (approved):
I like like like it!

contracts/BatchExchange.sol
```diff
@@ -239,7 +240,7 @@ contract BatchExchange is EpochTokenLocker {
     ) public returns (uint256) {
         require(acceptingSolutions(batchIndex), "Solutions are no longer accepted for this batch");
         require(
-            claimedObjectiveValue.mul(IMPROVEMENT_DENOMINATOR) > getCurrentObjectiveValue().mul(IMPROVEMENT_DENOMINATOR + 1),
+            isObjectiveValueGoodEnough(claimedObjectiveValue),
```
```
require(isFunctionNameGoodEnough, "Function name could probably be improved");
```

bh2smith (commented):
Looking pretty good. I would be marginally happier if that `isObjectiveValueGoodEnough` could be renamed... but have not been able to come up with anything all that good enough either.

- `sufficientObjective`
- `sufficentlyImproved`

I looked up some antonyms of marginal if this helps:

`vast`, `gross`, `non-negligable`, `significant`

Maybe `significantImprovement`

contracts/BatchExchange.sol
nlordell: Hehehe, I laughed out loud.

nlordell:
> Looking pretty good. I would be marginally happier if that `isObjectiveValueGoodEnough` could be renamed... but have not been able to come up with anything all that good enough either.

How about `isObjectiveValueSufficientlyImproved`?

bh2smith:
> How about isObjectiveValueSufficientlyImproved?

The thing I like the least is the `is`. Can we do this at all without it? I guess I'm just OCD... Maybe its not a problem.

fleupold:
I like the is because it makes clear what to expect from this method (a boolean answer in return). Other names such as check make me think it might just assert certain things. But no strong feelings, and don't want to filibuster this PR.

Should this go in the initial mainnet release or are we fine without it?

bh2smith:
Its fine as is... Please note that I certainly would **not** prefer something like `check`.

nlordell:
I wasn't a fan of "goodEnough" so i changed it to "sufficientlyImproved". I am also a fan of the `is` (and other similar verbs like `has`) as like @fleupold says it implies that it returns a boolean value.

bh2smith:
Thanks for taking care of this @nlordell!

PR #379, opened 2019-12-06
[Easy] Reordering variable declarations

josojo:
Anxo's comment about the order of variables and events:

I changed the ordering, such that constant variables are declared first, then they are followed by mutable variables and events


bh2smith:
> Anxo's comment about the order of variables and events

What was the comment?

bh2smith (approved):
Would love to see a better description here... Like did you make things alphabetical? or just reorder them? Not sure.

fleupold (approved):

PR #381, opened 2019-12-06
[Easy] Always use batchId instead of stateIndex

josojo:
One of anxo's comments


bh2smith (approved):
Good code changes. No comment on the description.

PR #396, opened 2019-12-12
[fix] add_token script

josojo:
small fix of import of the add_token script

testplan:

add an arbitrary token to the rinkeby exchange.


nlordell (approved):
Nice catch!

PR #397, opened 2019-12-12
Update of comment for disregarded utility calculation

josojo:
John W. found this one, see telegram chat


nlordell (approved):
Thanks John W.!

PR #399, opened 2019-12-12
Using user in withdraw

josojo:
In the function withdraw, we used `msg.sender` and the variable `user`.

Bug found by Adam.

I changed the tests, so that they would spot it.

---

How could this happen? Analysis:
The relevant change was made in this PR:
https://github.com/gnosis/dex-contracts/pull/233

First, I implemented the fix only with msg.sender and without any use of the "user", it got approved. And then I did the merge with master. Unfortunately, I did not pay enough attention to this merge and... the "user" was introduced in withdraw, but not in the new lines of withdraw().

What can we learn about?
- being more careful with merges - definitively
- approving after merges? - this is not practical, I guess.

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -104,10 +104,10 @@ contract EpochTokenLocker {
             "withdraw was not registered previously"
         );
         require(
-            lastCreditBatchId[msg.sender][token] < getCurrentBatchId(),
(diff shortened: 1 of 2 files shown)


fleupold (approved):
It's a pretty classic example of a hard to prevent merge conflict. It's a shame that the test suite didn't catch it.

PR #400, opened 2019-12-13
Include contracts and scripts into npm package

anxolin:
The migration scripts are in the NPM package, but the contracts were not published.

This PR includes the contracts in it.


josojo (approved):
Actually, the contracts would not be needed to do the migration. All necessary information should be in the builds.
So it is a pity that truffle requires it.

nlordell:
Do we need to include solidity dependencies (like openzeppelin) as well?

fleupold:
> Do we need to include solidity dependencies (like openzeppelin) as well?

I guess those can be reproduced by running npm install (although not the exact versions)?

What exactly is the use case for getting the contract sources? Would it be enough to just publish the interface? E.g. the driver is not really Interested in the sources, all it wants is the ABIs and deployed addresses.

Is there a way to have multiple packages published (e.g. a minimal version and a version with sources and dependencies)?

fleupold:
Any idea why the "unpacked size" is >7MB (https://www.npmjs.com/package/@gnosis.pm/dex-contracts/v/0.1.1)

When I install it locally I only get 300 Bytes

```
ls -lh node_modules/\@gnosis.pm/
total 0
drwxr-xr-x  10 felixleupold  staff   320B Dec 13 17:41 dex-contracts
```

anxolin:
> Actually, the contracts would not be needed to do the migration... 

Yep, I guess you mean the source is also in the build/*.json, and it's truffe format, so they should be able to know how to load contracts in your dependencies and migrate them. Agree

> Do we need to include solidity dependencies (like openzeppelin) as well?

This is becoming tricky. If we want to allow to migrate, I think so, otherwise someone using this repo will have to install manually the specific version you are using. But if we add those, the npm package is not small anymore.

> Any idea why the "unpacked size" is >7MB

Those json files are big:
image

image

This PR adds 100 KB only, it was 7MB already:
https://www.npmjs.com/package/@gnosis.pm/dex-contracts/v/0.1.0

anxolin:
Before we merge this maybe it's good to review how we expect someone else to migrate our project:
I guess some options are
- Clone as independent project, install, migrate
- Add submodule, install, migrate (I don't like git submodules too much)
- Add dependencies to owl, open zepelin, and expect people to use the migration scripts
- any other....??

I don't see any awesome way to do it


How someone would create a contract that depends on BatchExchange and work in local ganache?

nlordell:
I think the intention was option 3. In that case we need to move the solidity library dependencies into `dependencies` and not `devDependencies`:
- ~`@gnosis.pm/mock-contract`~ not needed since it should only be used in tests.
- `@gnosis.pm/owl-token`
- `@gnosis.pm/solidity-data-structures`
- `openzeppelin-solidity`
- `solidity-bytes-utils`

nlordell:
Also, some other issues I thought about:
1) Regarding including the scripts: they themselves have dependencies that are required to run them. Specifically they rely on truffle, and possibly some other libraries. So while the scripts are useful, they would require adding even more dependencies that aren't so slim.
2) It looks like `solidity-bytes-util` pulls in `truffle-hdwallet-provider` and `@gnosis/solidity-data-structures` pulls in `truffle` so the number of transient dependencies is getting pretty big. Specifically, `truffle-hdwallet-provider` as a dependency isn't great as it requires the infamous `sha3`, `keccak`, `secp256k1` and `scrypt` packages which need to build C++ source and don't work with recent NodeJS versions. For you to have an idea with just the 4 contract dependencies:
```shell
$ time -p npm ci
real 32.56
user 31.67
sys 3.24
$ npm ls --parseable | wc -l
579
$ yarn install # yarn is a little better
Done in 19.15s.
```

fleupold:
Wouldn't it be possible to just expose the BatchExchange/TokenEpochLockers' public interfaces and keep the implementation details outside the npm package?

anxolin:
I understand the approach is just expose the minimum, no dependencies, and then leave some instructions on "manual npm installs" in case you intend to use the "migration scripts"

If this is the case, it works for me too. It's not perfect, but I consider more important to have a lightweight NPM package so we don't enforce anyone to install too much libraries they might not need

nlordell:
Should we close this PR then?

anxolin:
well, why? we still need to be able to migrate from other projects, don't we?
How is the official way to do that?

nlordell:
I thought that the contract `.json` artifact files were enough for `truffle migrate`. And if that is not the case, then I suggest that we remove the migration scripts from the npm package and require the project to be cloned to do a migration.

fleupold:
> I thought that the contract .json artifact files were enough for truffle migrate. 

Yes that seems to be the case. I tried

```
npx truffle compile
rm -rf contracts
npx truffle migrate
```

And it worked fine, deploying all contracts to rinkeby.

anxolin:
Your are doing that from the same project.
Can you give the exact steps, for a brand new external project, on how to migrate? I see there's several options, but I'm not 100% sure about the one you are going for. A readme with steps would be ideal.

Also, what I'm more interested, what is the official way to depend on this contracts. Create a solidity contract that uses the contracts.

nlordell:
~~We can add it to the `README`~~
~~It looks like its already there: https://github.com/gnosis/dex-contracts#deployment-process~~
I misread what you wrote, we can add it to the `README`

fleupold:
This PR is outdated. I created https://github.com/gnosis/dex-contracts/issues/449 to track that we need to document how other projects should depend on us.

PR #402, opened 2019-12-16
better orderIndexManagement

josojo:
In the current implementation different uint types were used for the `orderIndex`
If someone would have stored over 65k(2**16) orders, then these orders could not have been considered in the solution, as the orderIndexs of submitSolution only accepted uint16.

This Pr unifies the uint value and makes sure that not more than 2**16 orders could be submitted.

Also it renames orderId to orderIndex

### DIFF
--- a/contracts/BatchExchange.sol
+++ b/contracts/BatchExchange.sol
@@ -75,7 +75,7 @@ contract BatchExchange is EpochTokenLocker {
     struct TradeData {
         address owner;
         uint128 volume;
-        uint16 orderId;
+        uint16 orderIndex;
     }
 
     struct SolutionData {
@@ -89,7 +89,7 @@ contract BatchExchange is EpochTokenLocker {
 
     event OrderPlacement(
         address owner,
-        uint256 index,
+        uint16 index,
         uint16 buyToken,
         uint16 sellToken,
         uint32 validFrom,
@@ -110,11 +110,16 @@ contract BatchExchange is EpochTokenLocker {
 
     /** @dev Event emitted when a new trade is settled
      */
-    event Trade(address indexed owner, uint256 indexed orderIds, uint256 executedSellAmount, uint256 executedBuyAmount);
+    event Trade(address indexed owner, uint16 indexed orderIndex, uint256 executedSellAmount, uint256 executedBuyAmount);
 
     /** @dev Event emitted when an already exectued trade gets reverted
      */
-    event TradeReversion(address indexed owner, uint256 indexed orderIds, uint256 executedSellAmount, uint256 executedBuyAmount);
+    event TradeReversion(
+        address indexed owner,
+        uint16 indexed orderIndex,
+        uint256 executedSellAmount,
+        uint256 executedBuyAmount
+    );
 
     /** @dev Constructor determines exchange parameters
       * @param maxTokens The maximum number of tokens that can be listed.
@@ -154,10 +159,10 @@ contract BatchExchange is EpochTokenLocker {
     /** @dev A user facing function used to place limit sell orders in auction with expiry defined by batchId
       * @param buyToken id of token to be bought
       * @param sellToken id of token to be sold
-      * @param validUntil batchId represnting order's expiry
+      * @param validUntil batchId representing order's expiry
       * @param buyAmount relative minimum amount of requested buy amount
       * @param sellAmount maximum amount of sell token to be exchanged
-      * @return orderId as index of user's current orders
+      * @return orderIndex as index of user's current orders
       *
       * Emits an {OrderPlacement} event with all relevant order details.
       */
@@ -176,7 +181,7 @@ contract BatchExchange is EpochTokenLocker {
       * @param validUntils batchIds represnnting order's expiry
       * @param buyAmounts relative minimum amount of requested buy amounts
       * @param sellAmounts maximum amounts of sell token to be exchanged
-      * @return `orderIds` an array of indices in which `msg.sender`'s orders are included
+      * @return `orderIndices` an array of indices in which `msg.sender`'s orders are included
       *
       * Emits an {OrderPlacement} event with all relevant order details.
       */
@@ -187,10 +192,10 @@ contract BatchExchange is EpochTokenLocker {
         uint32[] memory validUntils,
         uint128[] memory buyAmounts,
         uint128[] memory sellAmounts
-    ) public returns (uint256[] memory orderIds) {
-        orderIds = new uint256[](buyTokens.length);
+    ) public returns (uint16[] memory orderIndices) {
+        orderIndices = new uint16[](buyTokens.length);
         for (uint256 i = 0; i < buyTokens.length; i++) {
-            orderIds[i] = placeOrderInternal(
+            orderIndices[i] = placeOrderInternal(
                 buyTokens[i],
                 sellTokens[i],
                 validFroms[i],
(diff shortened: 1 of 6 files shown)


bh2smith (approved):
LGTM as long as all the places where u256 occurred have been changed.

bh2smith (commented):
do we need a new unit test?

bh2smith:
```
Error: CompileError: /home/travis/build/gnosis/dex-contracts/coverageEnv/contracts/BatchExchange.sol:534:44: TypeError: Invalid type for argument in function call. Invalid implicit conversion from uint256 to uint16 requested.
                emit TradeReversion(owner, orderId, sellAmount, buyAmount);
```

fleupold:
Agreed, they should be unified.

The reason for using uint16 is mainly gas, right? Would it make sense to benchmark the impact and make a call as to whether we want to use u16 or u256?

josojo:
I don't think that we need a benchmark, as uint16 is for sure cheaper. Each orderId supplied to submitOrder must be stored on-chain for a possible reversion of the trades. Hence, we save quite some gas.

And the only disadvantage of using uint16 instead of uint256 is that users cannot store more than 65k orders per address onchain. I don't think that this is really any practical downside.

anxolin (approved):
Great!

Did you review also the tests and utils for this change?

Unrelated to this PR, but I see that we still have in several parts of the code `stablex`, `auctionIndex`, etc..

networks.json
```diff
@@ -96,7 +96,7 @@
             },
             {
               "indexed": true,
-              "name": "orderIndexes",
+              "name": "orderIds",
```
bh2smith: Haven't we concluded that this would be `Index`?

networks.json
```diff
@@ -325,7 +325,7 @@
...
```
bh2smith: If we revert to using Index, this should be `orderIndices` not `Indexes` since indices is plural of index.

networks.json
josojo: Yes, these things will be updated with the new deployment

networks.json
josojo: Both is considered good english:
https://www.nasdaq.com/articles/indexes-or-indices-whats-the-deal-2016-05-12

networks.json
anxolin: relevant discussion is going on here: https://github.com/gnosis/dex-contracts/pull/403

Let's make a decision and be coherent between these 2 PRs

networks.json
bh2smith: Although this reference is about a _stock index_ (cf. nasdaq) where we are referring to a mathematical index. I believe the last sentence in the article you reference here states the following explicitly:


 "Indices" is used when referring to mathematical, scientific and statistical contexts.

anxolin: You are the native one :) 
I might be wrong, but I've always seen indexes when using it for IT

contracts/BatchExchange.sol
```diff
@@ -438,7 +445,7 @@ contract BatchExchange is EpochTokenLocker {
                 usedAmount: 0
             })
         );
-        uint256 orderIndex = orders[msg.sender].length - 1;
+        uint16 orderIndex = (orders[msg.sender].length - 1).toUint16();
```
nlordell, line 448: Does this require a test?

nlordell (approved):
Looks great! Just one question in comments.

networks.json
nlordell: :+1: for indices, it sounds more natural to me as well.

bh2smith (changes requested):
Please use "indices"! From the last sentence of the first paragraph of the following link.

https://www.nasdaq.com/articles/indexes-or-indices-whats-the-deal-2016-05-12

nlordell:
Another link: https://www.merriam-webster.com/dictionary/index
plural usually *indices* : a number or symbol or expression (such as an exponent) associated with another to indicate a mathematical operation to be performed or to **indicate use or position in an arrangement** [such as an array].

contracts/BatchExchange.sol
josojo: We used the toUint library in many places and only tested it in one spot. I think it is not necessary

josojo:
I am very sorry that I was force pushing. But I did not see any other option, as the merge was impossible, due to the renamings with batchId

bh2smith (changes requested):
Ok. Perhaps my previous comments have not been entirely clear. I was **not** referring to the variable name when insisting the term "indices" be used.

It has already been internally agreed upon (with very good reason) that `orderIds` is and should remain the **variable name** and this is completely fine. 

I only insist (with extremely strong opinion) that we use the term indices when referring to the plural form of this variable name in text (such as doc strings).

I apologise if this has become annoying or inconvenient, but can we please make this happen. Some explicit examples are made inline below. At the end I would expect that there are no occurrences of the variable names `orderIndex`, `orderIndices` or the word "indexes"  anywhere within this project.

contracts/BatchExchange.sol
```diff
@@ -154,10 +159,10 @@ contract BatchExchange is EpochTokenLocker {
     /** @dev A user facing function used to place limit sell orders in auction with expiry defined by batchId
       * @param buyToken id of token to be bought
       * @param sellToken id of token to be sold
-      * @param validUntil batchId represnting order's expiry
+      * @param validUntil batchId representing order's expiry
       * @param buyAmount relative minimum amount of requested buy amount
       * @param sellAmount maximum amount of sell token to be exchanged
-      * @return orderId as index of user's current orders
+      * @return orderIndex as index of user's current orders
```
bh2smith, line 165: Example:
```suggestion
      * @return orderId as index of user's current orders
```

contracts/BatchExchange.sol
```diff
@@ -241,16 +246,16 @@ contract BatchExchange is EpochTokenLocker {
...
         cancelOrders(cancellations);
         return placeValidFromOrders(buyTokens, sellTokens, validFroms, validUntils, buyAmounts, sellAmounts);
     }
 
     /** @dev a solver facing function called for auction settlement
       * @param batchId index of auction solution is referring to
       * @param owners array of addresses corresponding to touched orders
-      * @param orderIds array of order ids used in parallel with owners to identify touched order
-      * @param buyVolumes executed buy amounts for each order identified by index of owner-orderId arrays
+      * @param orderIndices array of order ids used in parallel with owners to identify touched order
```
```suggestion
      * @param orderIds an array of order indices used in parallel with owners to identify touched order
```

networks.json
bh2smith: @anxo - you have been reading poorly written IT documentation.

contracts/BatchExchange.sol
josojo: Sorry, but for me it makes no sense to call it `Ids` in the name and then `indices` the comments.

But good catch, some comments were not yet updated

josojo:
@bh2smith 
When I did the surgery/voting in dex-execution, I had indeed first mentioned orderId. But then @nlordell  brought up the compelling argument that we want orderIndex, hence I changed it and I got many more thumbs ups. 
So I guess it is discussed internally as `orderIndex`

nlordell:
I'm not strongly for either of them, and just slightly prefer `orderIndex`. I think the most important thing is consistency.

bh2smith:
~~I really do not want to see a variable called ‘orderIndices’ the thumbs up were for the original comment. Nobody thumbs upped the thread comments.~~

I guess it’s fine either way. I guess.

josojo:
This PR is blocking other work that I want to do. Hence, I will merge it now.
If there is more controversy, I will just open another PR.

bh2smith:
Please merge it. I will make the appropriate changes myself in a separate PR

contracts/BatchExchange.sol
bh2smith: variable name `somethingIds` followed by a description of what the variable is (i.e. indices of an array) has no conceptual conflict. In fact, I would argue that it makes the description more meaningful and rich. In the same respect that one would not define a term with the term itself.

For example, I consider it redundant to say the following:

"`orderIndex` is the index of an order"

PR #411, opened 2019-12-19
Simplifying the middleValue calculation in binary search

josojo:
Notice that this change is no equivalent for all inputs.
```
We found an example where (l + r) / 2 != l + (r - l) / 2

counterexample when (l, r) = (1, 0)
```

But in our situation with l<r, these values are the same, which can be proven by induction

### DIFF
--- a/contracts/libraries/TokenConservation.sol
+++ b/contracts/libraries/TokenConservation.sol
@@ -83,7 +83,7 @@ library TokenConservation {
         uint256 leftValue = 0;
         uint256 rightValue = tokenIdsForPrice.length - 1;
         while (rightValue >= leftValue) {
-            uint256 middleValue = leftValue + (rightValue - leftValue) / 2;
+            uint256 middleValue = (leftValue + rightValue) / 2;
             if (tokenIdsForPrice[middleValue] == tokenId) {
                 // shifted one to the right to account for fee token at index 0
                 return middleValue + 1;


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FchLfoLEerzXnG%2Fgiphy-downsized-medium.gif&token=simple_math" data-gitmeme-token="simple_math"></a>

bh2smith (approved):
Fantastic! The way binary search was meant to be!

contracts/libraries/TokenConservation.sol
```diff
@@ -83,7 +83,7 @@ library TokenConservation {
         uint256 leftValue = 0;
         uint256 rightValue = tokenIdsForPrice.length - 1;
         while (rightValue >= leftValue) {
-            uint256 middleValue = leftValue + (rightValue - leftValue) / 2;
+            uint256 middleValue = (leftValue + rightValue) / 2;
             if (tokenIdsForPrice[middleValue] == tokenId) {
                 // shifted one to the right to account for fee token at index 0
                 return middleValue + 1;
```
fedgiac, line 89: Looking at the change in context, this function should:
@return `index` in `tokenIdsForPrice` where `tokenId` appears (reverts if not found).
but it returns the index plus one. Maybe the documentation line should be updated?

fedgiac (commented):
Looks good!

PR #412, opened 2019-12-19
orderIndex -> orderId

bh2smith:
As discussed, agreed upon and grammatically correct this PR introduces the following changes:

1. Using `orderId` and `orderIds` in place of, unnecessarily and inconveniently long, variable names `orderIndex` and `orderIndices` respectively.

2. Ensuring no occurrences of the term "indexes" to refer to pluralized index of array elements.


fleupold (approved):
Maybe double-check with @josojo that he is on board.

contracts/BatchExchange.sol
```diff
@@ -212,7 +207,7 @@ contract BatchExchange is EpochTokenLocker {
       *
       * @param indices referencing the index of user's order to be canceled
       *
-      * Emits an {OrderCancelation} or {OrderDeletion} with sender's address and orderIndex
+      * Emits an {OrderCancelation} or {OrderDeletion} with sender's address and orderId
       */
     function cancelOrders(uint256[] memory indices) public {
         for (uint256 i = 0; i < indices.length; i++) {
```
nlordell: `indices` should be a `uint16[]` and follow the same naming convention as the other places.

nlordell (approved):
I'm OK with `orderId` but I do prefer `orderIndex` as these aren't really ids but indices in an array of orders. The reason id is a bit of a misnomer is that it an `orderId` doesn't uniquely identify an order and is dependent on the account. So `orderId == 1` will be different orders depending on what `msg.sender` is.

contracts/BatchExchange.sol
bh2smith: Are we sure this doesn't need to be u256 becaus eits going to be used in an array access?

bh2smith: It should be ID anyway right?

nlordell: We do array access with `uint16` in other places so I imagine its OK. Also it should be `orderIds` (or `orderIndices` if you agree with my comment :wink:).

contracts/BatchExchange.sol
```diff
@@ -162,7 +157,7 @@ contract BatchExchange is EpochTokenLocker {
       * @param validUntil batchId representing order's expiry
       * @param buyAmount relative minimum amount of requested buy amount
       * @param sellAmount maximum amount of sell token to be exchanged
-      * @return orderIndex as index of user's current orders
+      * @return orderId as index of user's current orders
```
```suggestion
      * @return orderId defined as the index in user's order array
```

contracts/BatchExchange.sol
bh2smith: I'll use `orderIds`.  I can see where you are coming from, and I do really like this suggestion (since it is not truly a unique identifier...) but I somewhat prefer the conciseness of orderIds.

josojo (approved):
good Feedback session :)

PR #413, opened 2019-12-19
Bump util-contracts to v2.0.6

bh2smith:
The following PR removes the cluttered event data from the `networks.json` file for easier visual detection of deployed contract addresses. Usually this file would be auto-generated by running the command `yarn run networks-extract` however, my computer kept changing the addresses (for some reason... @josojo  or @anxolin?) so I had to remove events manually for this PR.


anxolin (approved):
cool! it looks like some lines were killed

image

PR #414, opened 2019-12-19
[Easy] feeDenominator a contract constant

bh2smith:
Since the documentation for the product explicitly state that the fee is 0.1%  and constants are more gas efficient, the dfusion team has agreed to make this a constant with the contract. Changes included here are esssentially;

- make `feeDenominator` constant (equal to 1000) an rename as `FEE_DENOMINATOR`
- remove this parameter from the contract's constructor
- and adjust all relevant places that used to call the constructor (such as migrations)


josojo (approved):
NICE... thanks!

PR #416, opened 2019-12-19
355/lint warnings

### LINKED ISSUE #355: Address Lint Warnings
The CI is currently spitting out lint warnings, we should either fix them or ignore them:
```
> solhint "contracts/**/*.sol"

contracts/BatchExchange.sol
  14:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/DevDependencies.sol
  12:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
  12:1  warning  Code contains empty block                                 no-empty-blocks
contracts/EpochTokenLocker.sol
  14:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/libraries/SnappBaseCore.sol
  8:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/libraries/TokenConservation.sol
  9:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/SnappAuction.sol
  6:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/SnappBase.sol
  7:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/test/EpochTokenLockerTestInterface.sol
  7:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
contracts/test/TokenConservationWrapper.sol
  5:1  warning  Definition must be surrounded with two blank line indent  two-lines-top-level-separator
✖ 10 problems (0 errors, 10 warnings)
The command "npm run solhint" exited with 0.

$ solium -d contracts/

[Warning] No '.soliumignore' found. Use --no-soliumignore to make this warning go away.
contracts/BatchExchange.sol
  566:0    warning    Only use indent of 8 spaces.    indentation
contracts/DevDependencies.sol
  12:25    warning    Code contains empty block    no-empty-blocks
✖ 2 warnings found.
The command "solium -d contracts/" exited with 0.
```

bh2smith:
Closes #355 

In brief, this PR simply disables a few (extremely low severity) warnings such as having two lines at the top of every file... in order for the linters (solhint and solium) to fall in line better with prettier.


.solhint.json
```diff
@@ -7,9 +7,9 @@
     "max-states-count": ["error", 18],
     "compiler-fixed": true,
     "no-simple-event-func-name": true,
-    "not-rely-on-time": false,
+    "not-rely-on-time": "off",
```
nlordell: Question, does `not-rely-on-time` lint mean you can't use `now`? If so does it make more sense to just disable it with `/* solhint-disable not-rely-on-time */` or however you do it where we use `now`?

.soliumrc.json
```diff
@@ -6,6 +6,7 @@
   "rules": {
     "security/no-block-members": "off",
     "security/no-inline-assembly": "off",
+    "no-empty-blocks": "off",
```
nlordell, line 9: Same here, does it make more sense to disable the lint locally where this happens?

.solhint.json
bh2smith: Yes. I suppose. Good idea I will do this!

.soliumrc.json
bh2smith: Well ya. That is what I had before and I can easily make this the case again. Good catch.

.solhint.json
bh2smith: Only problem is, this will be A LOT of inline disables :


Screenshot 2019-12-19 at 4 45 24 PM


At this point, I would say we essentially want to turn it off.

josojo: I don't have a strong opinion, but it puts more lines into the smart contract on top without any real benefit.

josojo (approved):
I like it!

.solhint.json
bh2smith: I would propose either of the following two options:


1. Leave it as is with "off" for now... although this is equivalent to not including it anywhere?

2. I could disable-entire-file (once) at the top of each of the SNAPP contracts, while only disabling per use in the (more relevant) StableX contracts. 


I think option 2 would probably be the better option. Also the number of usages in the SNAPP contracts suggests that we could reduce the number by introducing some helper functions. 

What do you guys think? @josojo and @nlordell ?

bh2smith: In fact my proposal number two can be implements with only 5 lines of new code.

.soliumrc.json
bh2smith: In fact, I don't even think we need to worry about empty blocks anyway so I would argue that it be removed from solhint completely... (since off is the same as not mentioned)

.solhint.json
nlordell: Lets just globally disable it as you had before. Sorry for the inconvenience, I didn't expect it to be in so many places.

nlordell: If you like option 2, then go with that, I also think its better.

nlordell (approved):

.solhint.json
bh2smith: DUDE! I already reset hard and force pushed!  Good think I am quick!

nlordell: I'm good either way. I hadn't seen the messages after the screenshot because of Github things.

What I meant is that my initial suggestion of locally on each line disabling the use of `now` is invalid.

PR #418, opened 2019-12-19
improved tokenConservation array initialization

josojo:
Adams feedback:
why pass array to `TokenConservation.init`?

### DIFF
--- a/contracts/libraries/TokenConservation.sol
+++ b/contracts/libraries/TokenConservation.sol
@@ -10,10 +10,10 @@ library TokenConservation {
     using SignedSafeMath for int256;
 
     /** @dev initialize the token conservation data structure
-      * @param tokenIdsForPrice sorted list of tokenIds for which token conservation should be checked
+      * @param tokenNumber is the number of tokens the conservation should be checked for
       */
-    function init(uint256 size) internal pure returns (int256[] memory) {
-        return new int256[](size);
+    function init(uint256 tokenNumber) internal pure returns (int256[] memory) {
+        return new int256[](tokenNumber);
     }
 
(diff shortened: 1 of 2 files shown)


contracts/libraries/TokenConservation.sol
```diff
@@ -10,10 +10,10 @@ library TokenConservation {
     using SignedSafeMath for int256;
 
     /** @dev initialize the token conservation data structure
-      * @param tokenIdsForPrice sorted list of tokenIds for which token conservation should be checked
+      * @param tokenNumber is the number of tokens the conservation should be checked for
       */
-    function init(uint16[] memory tokenIdsForPrice) internal pure returns (int256[] memory) {
-        return new int256[](tokenIdsForPrice.length + 1);
+    function init(uint256 tokenNumber) internal pure returns (int256[] memory) {
```
nlordell, line 15: `tokenNumber` implies its the number of a single token, I think `tokenCount` is more appropriate here.

nlordell (commented):
I think we changed passing in the array so that the logic for how token conservation data is created is encapsulated in `TokenConservation` library. Also, from what I understand this is a delegate call anyways so I don't expect it to significantly impact the gas cost.

fleupold (approved):
Agree with Nick, the reason was an architectural decision. Not sure if it makes a difference gas wise (but if it does we can make this change). 

In general it would be nice if you could give a bit more context in the PR descriptions.

josojo:
Just tested it, no gas cost difference. Interesting.

Yeah, @nlordell , you analysis is correct besides the fact that we don't have delegate calls at all, as all functions are internal in the library.

PR #427, opened 2019-12-23
Include ERC20 Abi in NPM package

anxolin:
This PR includes the truffle artifact for IERC20 in the NPM package

It is convenient for us to get a IERC20 Abi in the NPM package so we don't need to include any additional dependency for that or to copy the file in out project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "build/contracts/SnappBase.json",
     "build/contracts/SnappBaseCore.json",
     "build/contracts/TokenConservation.json",
+    "build/contracts/ERC20.json",
     "networks.json",
     "src"
   ],
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "build/contracts/SnappBase.json",
     "build/contracts/SnappBaseCore.json",
     "build/contracts/TokenConservation.json",
-    "build/contracts/ERC20.json",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "build/contracts/SnappBase.json",
     "build/contracts/SnappBaseCore.json",
     "build/contracts/TokenConservation.json",
+    "build/contracts/ERC20.json",
```
nlordell: I think its better to include the `IERC20` interface here instead, this will also keep the package smaller as it has no bytecode and has the same ABI.

anxolin: good point.
good for me!

nlordell (approved):
Looks good to me!

nlordell:
@anxolin do you think including `@openzeppelin/contracts` is not a good idea? Its a pretty slim package without dependencies that just contains think like the IERC20/ERC20 truffle artifact.

anxolin:
> @anxolin do you think including @openzeppelin/contracts is not a good idea? Its a pretty slim package without dependencies that just contains think like the IERC20/ERC20 truffle artifact.

It's not a bad idea, although I would prefer to skip it for now since I'd like to keep `dex-js` dependencies as little as possible

bh2smith (approved):
Seems legit.

PR #434, opened 2019-12-31
Emit new token event

anxolin:
Proposal for adding a new event for listing tokens

The reasoning:
- It's the only important operation that doesn't have event
- An event doesn't add too much cost
- Doesn't affect to the cost of trading nor submit solutions
- Makes it easier for people to list the tokens by using web3 only
- Makes it easier to index in the graph for ganache and rinkeby, which doesn't support call handlers

### DIFF
--- a/contracts/BatchExchange.sol
+++ b/contracts/BatchExchange.sol
@@ -98,6 +98,8 @@ contract BatchExchange is EpochTokenLocker {
         uint128 priceDenominator
     );
 
+    event Token(address token, uint256 id);
+
     /** @dev Event emitted when an order is cancelled but still valid in the batch that is
      * currently being solved. It remains in storage but will not be tradable in any future
      * batch to be solved.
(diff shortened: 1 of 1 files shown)


contracts/BatchExchange.sol
```diff
@@ -146,6 +148,7 @@ contract BatchExchange is EpochTokenLocker {
             feeToken.burnOWL(msg.sender, FEE_FOR_LISTING_TOKEN_IN_OWL);
         }
         require(IdToAddressBiMap.insert(registeredTokens, numTokens, token), "Token already registered");
+        emit Token(token, numTokens);
```
josojo: Maybe we can call it `TokenListing`

josojo (approved):
We decided in the past against it, as people could just query all the listed tokens by asking the `tokenIdToAddressMap` with the numbers(0,1,2,3,4,5,...), 

But, if you say that there is software other would like to use, that does not support these call handlers, then I am okay with it.

contracts/BatchExchange.sol
bh2smith: Great name suggestion!

bh2smith (approved):
Wicked!

nlordell (approved):

PR #442, opened 2020-01-06
[Easy] OrderIds are u16

bh2smith:
An easy mistake (since `orderId` corresponds to an array index) is that `orderId` is `uint256`. However, for space saving reasons, we have decided to restrict this value to `uint16` (allows ~65K order per account.. which should suffice). 

**Test Plan:** Unit tests

### DIFF
--- a/contracts/BatchExchange.sol
+++ b/contracts/BatchExchange.sol
@@ -98,7 +98,7 @@ contract BatchExchange is EpochTokenLocker {
         uint128 priceDenominator
     );
 
-    event TokenListing(address token, uint256 id);
+    event TokenListing(address token, uint16 id);
 
     /** @dev Event emitted when an order is cancelled but still valid in the batch that is
      * currently being solved. It remains in storage but will not be tradable in any future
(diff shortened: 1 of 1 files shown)


contracts/BatchExchange.sol
```diff
@@ -233,7 +233,7 @@ contract BatchExchange is EpochTokenLocker {
       * @param sellAmounts maximum amounts of sell token to be exchanged in new orders
       * @return an array of indices in which `msg.sender`'s new orders are included
       *
-      * Emits {OrderCancelation} events for all cancelled orders and {OrderPlacement} events with all relevant new order details.
+      * Emits {OrderCancellation} events for all cancelled orders and {OrderPlacement} events with relevant new order details.
```
nlordell: Personally, I always go for Canadian/British spelling, so I prefer "cancelled". If we want to use American spelling, then we should change all instances of "cancelled" to "canceled" as well. Note that this change should be in a separate PR if we decide to do the change.

nlordell (approved):

contracts/BatchExchange.sol
bh2smith: I also prefer this in some situations. However I prefer to use a Z in words like realization and stuff like that.

bh2smith: Yes... I agree separate PR... I wish I had made more specific commits.

fleupold (approved):
🍨

anxolin:
Cool, thanks for the fast response @bh2smith

PR #471, opened 2020-01-16
Gas optimisation in updateupdateDepositsBalance

josojo:
HUGE gas optimization:

If I am running Ben's advanced "Large Examples" test-case without this optimization, it consumes 4264249  gas for the first submission and 5209727 gas for the second one.

With this gas optimization, we get:
3569261 gas for the first submission and 3846841 gas for the second one.

---
testplan: unit tests

### DIFF
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -214,6 +214,7 @@ contract EpochTokenLocker {
     function updateDepositsBalance(address user, address token) private {
         uint256 batchId = balanceStates[user][token].pendingDeposits.batchId;
         if (batchId > 0 && batchId < getCurrentBatchId()) {
+            // batchId > 0 is checked in order save an SSTORE in case there is no pending deposit
             balanceStates[user][token].balance = balanceStates[user][token].balance.add(
                 balanceStates[user][token].pendingDeposits.amount
             );
--- a/contracts/EpochTokenLocker.sol
+++ b/contracts/EpochTokenLocker.sol
@@ -212,7 +212,8 @@ contract EpochTokenLocker {
     }
 
     function updateDepositsBalance(address user, address token) private {
-        if (balanceStates[user][token].pendingDeposits.batchId < getCurrentBatchId()) {
+        uint256 batchId = balanceStates[user][token].pendingDeposits.batchId;
+        if (batchId > 0 && batchId < getCurrentBatchId()) {
             balanceStates[user][token].balance = balanceStates[user][token].balance.add(
                 balanceStates[user][token].pendingDeposits.amount
             );


josojo:
Should we raise `MAX_TOUCHED_ORDERS`, in order to get back to 5 M gas for the worst-case scenario?

contracts/EpochTokenLocker.sol
```diff
@@ -212,7 +212,8 @@ contract EpochTokenLocker {
     }
 
     function updateDepositsBalance(address user, address token) private {
-        if (balanceStates[user][token].pendingDeposits.batchId < getCurrentBatchId()) {
+        uint256 batchId = balanceStates[user][token].pendingDeposits.batchId;
```
fleupold, line 215: Maybe a comment that we do this to save an SSTORE in case there is no pending deposit would be appropriate?

fleupold (approved):
Nice find. I think there might be a bunch more optimizations like this hidden in the codebase. I think we can increase the constant to something around 5M gas in the worst case. Note, that this involves updating the extreme case example as well (but it could happen in parallel to the review).

nlordell (approved):
Great catch! All the virtual Eth animals will be happy that there are less emissions because of these gas savings.

josojo:
> Note, that this involves updating the extreme case example as well (but it could happen in parallel to the review).

@bh2smith you could do this the quickest. Do you want to do it?

bh2smith:
> @bh2smith you could do this the quickest. Do you want to do it?

Sure!

bh2smith (approved):
So... accessing the storage twice is a huge gas savings?

bh2smith:
> @bh2smith you could do this the quickest. Do you want to do it?

Sure!

bh2smith:
@fleupold 

> Note, that this involves updating the extreme case example as well (but it could happen in parallel to the review).

For what reason and what cases will this require a change to our examples? Are we going to increase max tokens? or what?

josojo:
> @fleupold
> 
> > Note, that this involves updating the extreme case example as well (but it could happen in parallel to the review).
> 
> For what reason and what cases will this require a change to our examples? Are we going to increase max tokens? or what?

Yes, we want to increase the max tokens. If you could figure out, which number would be the best, that would be awesome

josojo:
> So... accessing the storage twice is a huge gas savings?

The gas savings are coming from the fact that we do not write to storage if it is no needed. The newly introduced if statement checks, whether the storage update is really necessary, or whether it would just write the same number into storage.

bh2smith:
RIght yes, thanks for that. I had only seen that you were assigning a new variable, but didn't notice the change in the if condition. Is this reset to zero whenever there are no pendingDeposits? Can the same be done for withdraw?

josojo:
I checked the codebase and I can only find more point, where a similar SStore could be saved:
```
function addBalanceAndBlockWithdrawForThisBatch(address user, address token, uint256 amount) internal {
        if (hasValidWithdrawRequest(user, token)) {
            lastCreditBatchId[user][token] = getCurrentBatchId();
        }
        addBalance(user, token, amount);
    }
```

-->
```
function addBalanceAndBlockWithdrawForThisBatch(address user, address token, uint256 amount) internal {
        if (hasValidWithdrawRequest(user, token) &&  lastCreditBatchId[user][token] != getCurrentBatchId()) {
            lastCreditBatchId[user][token] = getCurrentBatchId();
        }
        addBalance(user, token, amount);
    }
```

This change would increase the gas costs (50 gas per order) in the majority of cases, since the if statement got more complex. And it would benefit us in the event that we really have a competing solution (450 gas per order).
To be honest, I don't understand the gas values, but I got them from the empiric evaluation of running the test:
```
    it.only("ensures that a solution reversion can not be prevented by additional withdrawRequests", 
``` 
on the other PR.

I would say, we don't need this gas improvement. What are your thoughts?

fleupold:
>For what reason and what cases will this require a change to our examples? Are we going to increase max tokens? or what?

Not max tokens, but max orders touched per batch

nlordell:
Maybe increasing the `MAX_TOUCHED_ORDERS` can be a separate PR?

fleupold:
Actually, thinking about the `MAX_TOUCHED_ORDERS` part again, I'm not sure we can increase it. They are supposed to cover the worst case and in the worst case I could have a pending deposit in all my accounts. 

Maybe the follow up task should be to add those pending deposits to the test case so that we get back to a real worst case scenario.

bh2smith:
@fleupold - Why pending deposits and not pending withdraws? Should the worst case cover/handle both?

bh2smith:
I have just added an extra occurrence of 

```
await makeDeposits(batchExchange, accounts, longRingTrade.deposits)
```
after the `closeAuction` command and before the first solution submission. Did not get any higher gas costs. In fact, I got higher gas costs without the extra pending deposits. Weird... I have also tried making the deposit requests between solution submission and got the same thing. Results are as follows:



**With deposits after close auction and before first solution:**

Screenshot 2020-01-17 at 11 13 30 AM



**Without pending deposits:**

Screenshot 2020-01-17 at 11 14 03 AM

**WIth deposits placed between solution submission:**

Screenshot 2020-01-17 at 11 14 28 AM

NOTICE THE NUMBER OF CALLS TO `deposit` is 50 in the first and third cases, while it is only 25 in the middle one.


Further more, it appears it is the competing solution whose gas cost is the lower of the two calls to submit solution in the case where deposits are made in between.

josojo:
@bh2smith Thanks for running the analsyis

> Did not get any higher gas costs.

This is really counterintuitive. Would you mind sharing your code for the test with pending deposits?

bh2smith:
Unfortunately these changes don't exist anywhere on a branch, but it is really only a single line difference to the master branch as described in the following quoted text.

> I have just added an extra occurrence of 
> `await makeDeposits(batchExchange, accounts, longRingTrade.deposits)`
> after the closeAuction command and before the first solution submission.
>  I have also tried making the deposit requests between solution submission and got the same thing.

josojo:
>Unfortunately these changes don't exist anywhere on a branch, but it is really only a single line difference to the master branch as described in the following quoted text.

I see.
I think that without this addition of the one line, we have already had the worst case. In this case, we have pending deposits, the initial deposits.

> Did not get any higher gas costs. In fact, I got higher gas costs without the extra pending deposits.

When you are doing the deposits after the closing of the auction, then you are effectively crediting the first deposits into the balance-state. The new deposits will then not be credited during the solution submission, as they are not yet valid in the last - currently solved - auction.

bh2smith:
Right yes! Mystery solved.

PR #476, opened 2020-01-17
Increase MAX_TOUCHED_ORDERS to 30

bh2smith:
This change only increases the value of a constant in the smart contract and updates the "worst case unit test" for largest possible ring trade having assertions on the maximum gas consumed for solution submission.

It was suggested [here](https://github.com/gnosis/dex-contracts/pull/471#issuecomment-575563978) that having pending deposits could also yield an even worse case, but these experiments turned out to have the opposite effect than predicted. Might be worth while investigating why we actually got lower gas costs when pending deposits exist.


bh2smith:
Note that this change will also require a small change of the analogous constant in the solver codebase.

test/stablex/batch_exchange.js
```diff
@@ -2044,14 +2044,16 @@ contract("BatchExchange", async accounts => {
...
-      const batchExchange = await setupGenericStableX(25)
+      const batchExchange = await setupGenericStableX(30)
       const sixPointFiveMillion = 6500000
-      await makeDeposits(batchExchange, accounts, longRingTrade.deposits)
+
+      const tradeExample = largeRing30
+      await makeDeposits(batchExchange, accounts, tradeExample.deposits)
       const batchId = (await batchExchange.getCurrentBatchId.call()).toNumber()
-      const orderIds = await placeOrders(batchExchange, accounts, longRingTrade.orders, batchId + 1)
+      const orderIds = await placeOrders(batchExchange, accounts, tradeExample.orders, batchId + 1)
```
fleupold, line 2061: this should be using largeRing30, shouldn't it?

fleupold (changes requested):

test/stablex/batch_exchange.js
josojo: He is setting 
const tradeExample = largeRing30, 
so I think it is good.

fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FK0Hy2NwI8IXZK%2Fgiphy.gif&token=blind" data-gitmeme-token="blind"></a>

I see now. But WHY?

We also need to make sure that the second solution actually attributes deposits to the balances that are traded in it. So either the users or the tokens need to be different than in the first solution.

fleupold: We should also make sure that all users have a pending withdraw requests on all tokens that gets "blocked" when their balance increases to maximize the worst case scenario.

bh2smith: oh my... I will have to think about this a bit. I mean, I am not sure I understand why we would want to "Block" the trade from happening, if we are trying to successfully submit a solution...

bh2smith:
>  So either the users or the tokens need to be different than in the first solution.

The most recent commit introduces another 30 orders from users 2 and 3, with the first solution filling a ring for users 0 and 1 while second solution reverts and matches users 2 and 3's ring.

bh2smith:
>  So either the users or the tokens need to be different than in the first solution.

The most recent commit introduces another 30 orders from users 2 and 3, with the first solution filling a ring for users 0 and 1 while second solution reverts and matches users 2 and 3's ring.

bh2smith:
In response to @fleupold's suggestions that competing solutions should have either different users or different tokens and that there should exist withdraw requests.

I am fairly certain that having a competing solution using different users or different tokens would not make any difference with regards to gas consumption.

The steps taken within `submitSolution` are summarized as follows:

- validate solution
- evaluate and compare objective function
- revert previous solution
- apply new solution

**Notice that the last two steps occur independently** (regardless of users or tokens involved). I can only assume that your concern with our existing example not being the "true worst case" stems from an assumption that previous solutions are overwritten rather than reverted and rewritten (in which case, I would  totally agree).

At this point I am unable to determine the relevance of having withdraw requests made between solution submission's(or at any stage in this example) in our search for the true worst gas consumption.

I can admit that I have built this example from memory and will take another closer look at the contract code to see if there are any branches this example does not touch in order to maximize gas consumption!

josojo:
> At this point I am unable to determine the relevance of having withdraw requests made between solution submission's(or at any stage in this example) in our search for the true worst gas consumption.

We should make sure that in the worst case scenario for every account a withdrawal blocker is set, via 
https://github.com/gnosis/dex-contracts/blob/master/contracts/EpochTokenLocker.sol#L198,
while a new solution is submitted.

fleupold:
> I am fairly certain that having a competing solution using different users or different tokens would not make any difference with regards to gas consumption.

If you don't have different users/tokens, whenever you submit the second solution all deposits have already been applied (when submitting the first solution). Thus applying the second solution could consume more gas if we used completely new users/tokens and also had to apply their deposits.

bh2smith:
We now have 60 orders (first 30 for large ring trade involving user 0 and 1 and second with users 2 and 3). There are valid withdraw requests for all orders/accounts and gas reported locally is 

```
First submissions gas used: 4624954
Second submissions gas used: 4307440
```
This PR is ready for review. I am prepared to experiment with higher values for `MAX_TOUCHED_ORDERS`

I believe the gas used on second submission is lower because of the gas refund for reverted info, but not sure.

test/stablex/batch_exchange.js
```diff
@@ -2063,11 +2083,17 @@ contract("BatchExchange", async accounts => {
...
       assert(
-        firstSubmissionTX.receipt.gasUsed < sixPointFiveMillion,
-        `Solution submission exceeded 6.5 million gas at ${firstSubmissionTX.receipt.gasUsed}`
+        firstSubmissionTX.receipt.gasUsed < fivePointFiveMillion,
+        `Solution submission exceeded 5.5 million gas at ${firstSubmissionTX.receipt.gasUsed}`
       )
 
-      const solution2 = solutionSubmissionParams(longRingTrade.solutions[1], accounts, orderIds)
+      // Ensure second 30 order's users valid withdraw requests.
+      for (const deposit of tradeExample.deposits.slice(30)) {
```
fleupold, line 2091: Doesn't this have to happen before the order collection is ended? Otherwise `hasValidWithdrawRequest` will be false and thus we won't hit [this line](https://github.com/gnosis/dex-contracts/blob/2dfb10ea610e12109460c499c548361c6beb7ce1/contracts/EpochTokenLocker.sol#L198) in the second solution submission.
Thus it's not the worst case.

test/resources/examples/index.js
```diff
@@ -91,149 +91,162 @@ const basicRingTrade = generateTestCase({
...
-    { sellToken: 23, buyToken: 24, sellAmount: toETH(1), buyAmount: toETH(0.99), user: 1 },
-    { sellToken: 24, buyToken: 0, sellAmount: toETH(1), buyAmount: toETH(0.99), user: 0 },
-  ],
+  orders: [...Array(n).keys()]
+    .map(i => ({
+      sellToken: i % n,
+      buyToken: (i + 1) % n,
+      sellAmount: toETH(1),
+      buyAmount: toETH(0.99),
+      user: i % 2,
```
fleupold, line 103: nit: would it be easier to understand this test if we just used i as the user and let it run from 0 to 60?

test/stablex/batch_exchange.js
```diff
@@ -2044,14 +2044,34 @@ contract("BatchExchange", async accounts => {
...
       const batchId = (await batchExchange.getCurrentBatchId.call()).toNumber()
-      const orderIds = await placeOrders(batchExchange, accounts, longRingTrade.orders, batchId + 1)
+      const orderIds = await placeOrders(batchExchange, accounts, tradeExample.orders, batchId + 1)
       await closeAuction(batchExchange)
 
-      const solution = solutionSubmissionParams(longRingTrade.solutions[0], accounts, orderIds)
+      // Ensure that first 30 orders have valid withdraw requests.
+      for (const deposit of tradeExample.deposits.slice(0, 30)) {
+        const tokenAddress = await batchExchange.tokenIdToAddressMap.call(deposit.token)
+        assert(
```
fleupold, line 2067: How long is this test taking? I'm wondering if we should make all these calls here or if one check is sufficient?

fleupold (commented):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FrZ10JtoPRv1Ly%2Fgiphy.gif&token=almost" data-gitmeme-token="almost"></a>

test/resources/examples/index.js
bh2smith: There aren't enough accounts loaded into ganache for that.

test/stablex/batch_exchange.js
bh2smith: I don't understand the problem here. The orders have all been placed and all of these assertions are evaluating to `true`.

test/stablex/batch_exchange.js
bh2smith:
```
    Large Examples
      ✓ ensures hard gas limit on largest possible ring trade  (35710ms)
```

I could reduce this to one assertion.

fleupold: 35 seconds for a single test is really long. Maybe we can at least not sequentially await on all future but instead issue the requests in parallel? That is, use [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) instead of sequential awaits.

test/stablex/batch_exchange.js
fleupold: You are right. I though you were requesting it here, but you are just asserting.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FYl5aO3gdVfsQ0%2Fgiphy.gif&token=nothing_to_see" data-gitmeme-token="nothing_to_see"></a>

fleupold (approved):
Would be nice to `await Promis.all` whenever we have awaits in a for loop. But this is also something we can probably try to apply to the entire codebase, so could be done in a separate PR.

test/stablex/batch_exchange.js
bh2smith: I like this idea much better!

bh2smith:
> `await Promise.all` whenever we have awaits in a for loop. But this is also something we can probably try to apply to the entire code base, so could be done in a separate PR.

I like the separate PR idea and will create and issue. #483

test/stablex/batch_exchange.js
```diff
@@ -2044,14 +2044,34 @@ contract("BatchExchange", async accounts => {
...
+
       const batchId = (await batchExchange.getCurrentBatchId.call()).toNumber()
-      const orderIds = await placeOrders(batchExchange, accounts, longRingTrade.orders, batchId + 1)
+      const orderIds = await placeOrders(batchExchange, accounts, tradeExample.orders, batchId + 1)
       await closeAuction(batchExchange)
 
-      const solution = solutionSubmissionParams(longRingTrade.solutions[0], accounts, orderIds)
+      // Ensure that first 30 orders have valid withdraw requests.
+      for (const deposit of tradeExample.deposits.slice(0, 30)) {
+        const tokenAddress = await batchExchange.tokenIdToAddressMap.call(deposit.token)
```
josojo, line 2066: Here, you are checking that the withdrawRequest is available for the selltoken, but I guess we need to have it for the bought tokens.

bh2smith: Oh no. Really? Why? Nobody has a balance of buy token to withdraw anyway.

fleupold: You are right. The withdraw is only blocked when we add balance, so we need to request the withdraws for the buy not the sell token. 

Great catch.

fleupold: @bh2smith. They will after the first solution is applied. That’s why we need to block the withdraw because that balance is not yet final and thus shouldn’t be withdrawn until the next batch.

bh2smith: How can all this stuff happen in one batch. I will revert and try again

bh2smith: How can all this stuff happen in one batch. I will revert and try again

PR #485, opened 2020-01-21
Update Worst Case Trade

bh2smith:
In the large trade example, in order to have the worst gas consumption, we must make withdraw requests on the tokens acquired during trade rather than the sell tokens. First commit will likely result in a failed test and display the maximum gas required for trade. Will likely have to put the gas cap at 6.5M (as it was before).


josojo (approved):
Nice!!!

New gas cost is 6022271.

bh2smith:
> New gas cost is 6022271.

Yes, although this will be higher once the Event Emission is updated #479

PR #493, opened 2020-01-22
remove DEFAULT_GAS_LIMIT

Velenir:
Trying to remove DEFAULT_GAS_LIMIT
Without it **ganache** would default to `6721975`


bh2smith:
This will likely fail on our Large Test Example:

```
  Contract: BatchExchange
    Large Examples [ @skip-on-coverage ]
      1) ensures hard gas limit on largest possible ring trade 
```

PR #510, opened 2020-01-29
Deployment fix3

josojo:
This is roughly how Felix imagined solving the migration problem. 

But it turns out that libraries imported from npm modules can not easily be linked in deployment scripts.
I suggest to just publish the .sol files for the migration of external projects


src/migration/PoC_dfusion.js
```diff
@@ -27,28 +27,25 @@ async function migrate({ artifacts, deployer, network, account, web3, maxTokens
...
-    deployer,
-    account,
-    "@gnosis.pm/solidity-data-structures/build/contracts/IdToAddressBiMap"
-  )
-  const IterableAppendOnlySet = getDependency(
-    artifacts,
-    network,
+  const BiMap = artifactFromNpmImport("@gnosis.pm/solidity-data-structures/build/contracts/IdToAddressBiMap", deployer, account)
+  const biMap = await BiMap.deployed()
+  biMap.contract_name = "IdToAddressBiMap"
```
josojo, line 32: These field contract_name and isDeployed are expected, but they are not set by @truffle/contract

PR #533, opened 2020-02-10
[Very Easy] Don't need empty contract declaration!

bh2smith:
Just discovered that this DevDependencies contract doesn't need to have an empty block


josojo (approved):
Cool

PR #542, opened 2020-02-13
[Easy] Bump version to v0.2.2

nlordell:
This bumps contract version to v0.2.2. From what I understand we want to used released version with `dex-services` and the recent change #532 has the nice side effect of including `ERC20Detailed` contract in the build output which the price estimation wants to use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gnosis.pm/dex-contracts",
-  "version": "0.2.1",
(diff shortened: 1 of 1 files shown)


josojo (approved):
>e #532 has the nice side effect of including ERC20Detailed contract in the build output which the price estimation wants to use.

yeah, this is a really nice side effect  :)

PR #544, opened 2020-02-13
[Easy] Move and rename utility function

bh2smith:
`decodeAuctionElements` was in test utilities, but would be nice to expose for integration projects (like `dex-liquidity-provision`) to `decodeOrdersBN` which has the exact same functionality and can still be used for testing. 

Perhaps we would like to reevaluate what other helper methods should/could be exposed here.


nlordell (approved):
Love it!

fleupold (approved):

josojo (approved):

PR #583, opened 2020-03-12
Add thegraph data script

anxolin:
Adds  a new script file (`setup_thegraph_data.js`) with some basic data, useful for testing the graph.

Easier to read if you start in line 130. Prepares some test users, tokens, trades, apply some solution, do a cancelation and a deletion.


scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,233 @@
...
+    // Set user1 as minter of OWL
+    const owl = await _getOwl()
+    await owl.setMinter(user1.address)
+
+    // Mint owl for user1 and user2
+    const amount = web3.utils.toWei("3000")
+    await _mintOWL({ user: user1, owl, amount, batchExchange })
+    await _mintOWL({ user: user2, owl, amount, batchExchange })
+
+    // Create 3 tokens
```
josojo: Maybe you can adapt it here to 2 tokens and delete the comment

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,233 @@
...
+const { closeAuction } = require("../../scripts/stablex/utilities.js")
+const { decodeOrdersBN } = require("../../src/encoding")
+
+const { toETH, feeAdded, feeSubtracted, ERROR_EPSILON } = require("../../test/resources/math")
+const { waitForNSeconds } = require("../../test/utilities")
+
+const BatchExchange = artifacts.require("BatchExchange")
+const ERC20Mintable = artifacts.require("ERC20Mintable.sol")
+
+async function _getOwl() {
```
josojo: It would be good if you could put the helper functions into the utiltities.js file. There they are much easier to find and we can also use them in unit tests, hence refactoring errors are caught better.

Much of the OWL minting and token adding functionality could/should also be used from here:
https://github.com/gnosis/dex-contracts/blob/master/scripts/stablex/utilities.js#L38

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,233 @@
...
+  await token.instance.mint(user.address, amount)
+  await token.instance.approve(batchExchange.address, amount, { from: user.address })
+}
+
+async function _getBatchId(batchExchange) {
+  const batchId = await batchExchange.getCurrentBatchId.call()
+  return batchId.toNumber()
+}
+
+function _getTestCase({ user1, user2, token1 }) {
```
josojo: Do you want to use this function in the future in different ways? If not, if it is really only used once, you could just get the example directly from here
 https://github.com/gnosis/dex-contracts/blob/master/test/resources/examples/index.js#L7
and reduce code duplication.

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,233 @@
...
+    // Mint the 2 tokens for the 2 users
+    // const tokens = [owl, token1, token2, token3]
+    const tokens = [token1, token2]
+    tokens.forEach(token => {
+      _mintToken({ token, user: user1, amount, batchExchange })
+      _mintToken({ token, user: user2, amount, batchExchange })
+    })
+
+    // Generate a basic test case
+    const testCase = _getTestCase({ user1, user2, token1, token2 })
```
josojo: I think  token2 is not used in your example. You are only trading against owl

josojo (approved):
Cool, that script will totally help. Thanks for taking the initiative.

Just some comments about code reusage

scripts/stablex/setup_thegraph_data.js
anxolin: I didn't do it cause that code was relying on a clean newly deployed contract (no previous tokens had been listed, so we use token 0 and 1)

So in my approach I was planning to rely only on my data. Anyways, u are right. I'll try to reuse this cases.

scripts/stablex/setup_thegraph_data.js
anxolin: Okay, I'll move there more logic and refactor a bit

scripts/stablex/setup_thegraph_data.js
anxolin: Yes, u right. I was going to create another example, but for now didn't use it.
I'll remove it and add it when/if I evolve the script

test/stablex/stablex_utils.js
```diff
@@ -71,8 +70,29 @@ const placeOrders = async function(contract, accounts, orderList, auctionIndex)
   return orderIds
 }
 
+async function createMintableToken(artifacts) {
```
fleupold: I find it weird that some methods require the artefacts to be passed in and others don't. Maybe this is not the right place to add this method? If it's not used in unit tests it should probably not be in the test folder but instead in the top level utilities or script utilities.

scripts/stablex/utilities.js
```diff
@@ -143,14 +167,88 @@ const sendLiquidityOrders = async function(
...
+    }
+  }
+}
+
+async function setAllowance({ token, account, amount, batchExchange }) {
+  console.log("Set allowance of %d for token %s and user %s", amount, token.address, account)
+  await token.approve(batchExchange.address, amount, { from: account })
+}
+
+async function mintOwlForUsers({ users, amount, owl }) {
```
fleupold: This method implies that users[0] is the minter. I think it therefore either set the minter itself or take it as an argument.

josojo (approved):
Nice. 
Thanks for going the extra step and amending the addToken function!

scripts/stablex/utilities.js
```diff
@@ -35,34 +48,45 @@ const fetchTokenInfo = async function(
...
 
-const addTokens = async function(token_addresses, web3, artifacts) {
-  const accounts = await web3.eth.getAccounts()
-
-  const BatchExchange = artifacts.require("BatchExchange")
-  const instance = await BatchExchange.deployed()
-
-  const TokenOWL = artifacts.require("../node_modules/@gnosis.pm/owl-token/build/contracts/TokenOWL")
-  const owl = await TokenOWL.at(await instance.feeToken.call())
+const addTokens = async function({ tokenAddresses, account, batchExchange, owl }) {
```
josojo, line 56: I think your approach is the cleaner one: handing down the contract instances: batchExchange instead of the artifacts and the web3 object.

It is just that we have structured most other functions differently in the past. But since your approach is the "cleaner" one, lets stick with yours and amend the others over time

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,118 @@
...
+    })
+
+    // List the tokens in the exchange
+    const tokenAddresses = [token1Instance.address]
+    const [token1] = await addTokens({ tokenAddresses, account: user1Address, batchExchange, owl })
+
+    // Mint tokens for every user
+    await mintTokens({
+      tokens: tokensInstances,
+      users: [user1Address, user2Address],
```
fleupold: nit: reuse usersAddresses var?

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,118 @@
...
+    // Close the auction
+    console.log("Close auction so we can withdraw the tokens")
+    await closeAuction(batchExchange, web3)
+
+    // Withdraw tokens
+    console.log("Withdraw for user %s, token %s (%d)", user1Address, token1.address, token1.id)
+    await batchExchange.withdraw(user1Address, token1.address, { from: user1Address })
+
+    // Submit solution
+    await submitSolution({
```
fleupold, line 86: Did you want to use an example with a solution and a better solution to get revert events?

scripts/stablex/utilities.js
fleupold, line 56: I like the trick to get named arguments! 👍

scripts/stablex/utilities.js
```diff
@@ -143,14 +167,88 @@ const sendLiquidityOrders = async function(
...
   closeAuction,
   token_list_url,
   fetchTokenInfo,
   sendLiquidityOrders,
   getOrdersPaginated,
   maxUint32,
+  setAllowance,
+  setAllowances,
+  mintOwlForUsers,
+  mintOWL,
```
fleupold: Instead of exporting the singular and plural methods, would it make sense to just use the plural with a one element list? The more methods you export the more confusing it becomes.

fleupold (commented):
Looks good, some smaller comments inline. I really like the named arguments style.
Can you add a testplan that makes sure the affected existing scripts are still working (by running them locally)?

scripts/stablex/utilities.js
```diff
@@ -35,34 +48,45 @@ const fetchTokenInfo = async function(
...
 
-  const feeForAddingToken = (await instance.FEE_FOR_LISTING_TOKEN_IN_OWL.call()).mul(new BN(token_addresses.length))
-  const balanceOfOWL = await owl.balanceOf.call(accounts[0])
-  if (feeForAddingToken.gt(balanceOfOWL)) {
-    console.log("More fee tokens are required to add all tokens")
-    return
+  // Ensure the user has enough OWL balance
+  const balanceOfOWL = await owl.balanceOf.call(account)
+  if (totalFees.gt(balanceOfOWL)) {
+    throw new Error("More fee tokens are required to add all tokens")
```
anxolin, line 64: @fleupold @josojo I just want to double check that this is fine with you guys.
I hope I don't break anything with this. I just thought this is an error that should be raised instead of console logged.

scripts/stablex/utilities.js
anxolin: I'll add a minter parameter so we keep the two logics separated

scripts/stablex/setup_thegraph_data.js
anxolin: good catch

scripts/stablex/setup_thegraph_data.js
anxolin: It would be nice! it was in my plans. I would love to do that in another PR

scripts/stablex/utilities.js
anxolin: Here it is: 06385f38a58f676b5789dacf7e754300bbc05710

scripts/stablex/utilities.js
anxolin: okay, no problem. Removing singulars

anxolin: 47016fbf956b13bbbae3e5d98dc4bc78404b6e0d

test/stablex/stablex_utils.js
anxolin: I tried to reduce the use of `artifacts` to the minimun. This is why there's a single `getBatchExchange(artifacts)` and `getOwl(artifacts)` and then used named parameters to hand the functions the instance of the contracts.

The same applies to the creation of `ERC20Mintable`.  Anyways, I've moved it to were you suggested.

scripts/stablex/utilities.js
fleupold: Yes, I agree that's better.

test/stablex/stablex_utils.js
```diff
@@ -18,9 +15,11 @@ const { sendTxAndGetReturnValue } = require("../utilities")
  * @returns {}
  */
 const setupGenericStableX = async function(numTokens = 2, maxTokens = 2 ** 16 - 1) {
+  const MockContract = artifacts.require("MockContract")
```
fleupold, line 18: nit: revert? I'm not sure what the implications are for inline requiring vs. top-level, but it seems to be more common to do requires at the top if possible.

fleupold (approved):
Nice work! With this we might even be able to build some automated e2e tests for the graph repo to make sure all events are processed as we intend to.

test/stablex/stablex_utils.js
anxolin: Yes, I will revert that. I don't even remember changing it. 

This change made no sense I agree. Also, imo, doing it in the top is also not a good idea for a utils file. Although that's a refactor totally out of this PR. Doing it on top means you assume the JS file would be executed within a context, where you expect some magical global vars to be injected.

It's kind of fine for truffle scripts, as an entry point. But the utils shouldn't make any assumption, they should get all the parameters they need.

Some utils solve that by receiving the `artifcats` and pulling from there, but I personally believe, usually is better to be specific about the parameters and name the contracts you rely on.

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,122 @@
...
+
+module.exports = async function(callback) {
+  try {
+    const batchExchange = await getBatchExchange(artifacts)
+    const owl = await getOwl(artifacts)
+
+    // Prepare user accounts
+    const accounts = await web3.eth.getAccounts()
+    const [user1Address, user2Address] = accounts
+    const usersAddresses = [user1Address, user2Address]
```
W3stside, line 28: Why this? userAddress and accounts are the same thing no?

W3stside: you could condense lines `27, 28, 29` via `const [user1Address, user2Address] = await web3.eth.getAccounts()`

scripts/stablex/setup_thegraph_data.js
```diff
@@ -0,0 +1,122 @@
...
+const { toETH } = require("../../test/resources/math")
+const { waitForNSeconds } = require("../../test/utilities")
+
+module.exports = async function(callback) {
+  try {
+    const batchExchange = await getBatchExchange(artifacts)
+    const owl = await getOwl(artifacts)
+
+    // Prepare user accounts
+    const accounts = await web3.eth.getAccounts()
```
W3stside:
```suggestion
    const [user1Address, user2Address] = await web3.eth.getAccounts()
```

test/stablex/stablex_utils.js
anxolin: Okay @fleupold , I removed that and I broke my script. 

Now I see why I did the change. As I said this relies in a magic globals. I think someone had the same issue when created the `setup_environment` script, and it seems like he opt in to duplicate the logic from this `setupGenericStableX`

This setup function should receive either the artifacts or the 2 contracts and be refactorized across the project. Imo it should be the contracts.

The thing is this falls way out of this PR, as you see, this function is used a little and I don't want to overcomplicate this PR:

image

I think a review of the dependencies and some refactor would be beneficial in the project. Not sure if this is to ambitious, but ideal if this refactor would be used to do the typescript jump? Probably there's pseudo-automated tools already to migrate a whole project to TS.

anxolin: To be more specific about the error I was getting is:
* Utils file export some functions that are useful in some scripts
* Initially you would think that by requiring the file no code get's executed (no 
* The moment you require the utils file it's going to try to use `artifacts` that won't be available in that context

This is why this "assumptions" specially for utils is not good

scripts/stablex/setup_thegraph_data.js
anxolin: Good question. It looks the same, but is not :)

accounts has all the web3 accounts. This is for ganache so you'll see 10 accounts there.
I'm using destructuring to have a slice of that array while I create some consts for the individual  addresses that I will need later

test/stablex/stablex_utils.js
fleupold: My point is that since this file is located inside **test**/stablex/ it should only be used by tests and not by external scripts. If they are useful as top-level utils we should move them to src or some other "shareable".

If you feel this is outside the scope of this PR I can think three options:
1. Create a PR that just moves this function outside of its current location to a shareable one, then rebase this PR on top.
2. Copy the code into utils for now and create a follow up PR that merges them
3. Leave it as is and do a follow up PR extracting the method

anxolin: image

But this is reusing the test data, and those function `makeDeposits`, `placeOrders` are relying on the structure used in `test/resources/examples`

I feel it's OK using test data for a test setup in targeted for a local ganache. And since you have the utils for making the trades and deposits I've just used them (Alex gave me the tip)

I also use other tools that are in the test utilities, like `waitForNSeconds`, since it's targeted for ganache setup.

What's would be your concern?

I can move the script to the test folder if u think it's better

fleupold: It's a good point that `waitForNSeconds` is also used in other scripts and comes from the test utilities.
I still think it would be cleaner to lift scripts that are used for both unit tests as well as scripts into a top-level shared utility (instead of cross referencing scripting and unit tests dependencies). But since this is not yet the case it doesn't have to be changed here.

I think the script should definitely be part of the scripts subfolder.

🚢

PR #699, opened 2020-04-29
Bump contracts to 0.2.12 and deploy

nlordell:
:point_up: 

### Test Plan

Check out the newly verified and deployed BatchExchangeViewer contract on [Etherscan.io](https://etherscan.io/address/0x1175AFf51996ADBE45a1c330c0B999D104016ba9#code)

### DIFF
--- a/networks.json
+++ b/networks.json
@@ -20,13 +20,13 @@
   "BatchExchangeViewer": {
     "1": {
       "links": {},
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Great! We could probably make the verification step part of travis also so we don't have to do it manually.

nlordell:
> We could probably make the verification step part of travis also so we don't have to do it manually.

Sounds like s great idea!

PR #700, opened 2020-04-29
[Easy] Add offset to target buffer

nlordell:
#698 introduced a bug where the target buffer was getting zeroed out in the first position and not in the position where the element was being encoded.

### Test Plan

Checkout `d6414f7` and see the test fail with order ID 5 not getting correctly filtered (and having duplicate data to order ID 2)

Checkout `fix_buffer_zeroing` and see the test now succeeds.


fleupold (approved):
🦅

nlordell:
:see_no_evil:

PR #703, opened 2020-04-30
[Events Orderbook] Query all exchange events. (1)

nlordell:
This PR was moved from gnosis/dex-price-estimator#23

This PR implements event querying in pages for the batch exchange contract that will eventually be used to build an event based orderbook.

Currently the events aren't being used for anything, but there are a few assertions that were added to the event processing to ensure that some basic assumptions are indeed correct. These assumptions will be used in order to implement account state updating:
- Events from past blocks are never applied retro-actively
- Events are always in order

Furthermore some basic boilerplate code is setup in order to get a contract instance and a web3 provider.

### Test Plan

Run the code from the new module and see it fetch all events since the dawn of the dinosaurs and have no assertions fail.
```
export ETHEREUM_NODE_URL=... or export INFURA_PROJECT_ID=...
yarn test-streamed-orderbook 
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,14 +3,14 @@ module.exports = {
   env: {
     browser: true,
     commonjs: true,
-    es6: true,
+    es2020: true,
   },
   extends: "eslint:recommended",
   parserOptions: {
     ecmaFeatures: {
       jsx: true,
     },
-    ecmaVersion: 2018,
+    ecmaVersion: 2015,
     sourceType: "module",
   },
   plugins: ["react", "@typescript-eslint"],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,5 @@
 module.exports = {
-  parser: '@typescript-eslint/parser',
+  parser: "@typescript-eslint/parser",
   env: {
     browser: true,
     commonjs: true,
@@ -13,10 +13,7 @@ module.exports = {
     ecmaVersion: 2018,
     sourceType: "module",
   },
-  plugins: [
-    "react",
-    "@typescript-eslint",
-  ],
+  plugins: ["react", "@typescript-eslint"],
   rules: {
     indent: ["error", 2],
     "linebreak-style": ["error", "unix"],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,5 @@
 module.exports = {
+  parser: '@typescript-eslint/parser',
   env: {
     browser: true,
     commonjs: true,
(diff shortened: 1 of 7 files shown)


.eslintrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
```
e00E, line 1: Is this intended to be in this PR. I don't have a problem with it, just want to point it out.

e00E (approved):
This has the same contents as the linked and already reviewed PR, right?

.eslintrc.js
nlordell: I opened #702 to fix these issues, so some of these changes will disappear when I merge or rebase master.

nlordell:
> This has the same contents as the linked and already reviewed PR, right?

There are a couple differences: We no longer split the events into batches and the test plan is done with a proper Mocha test. To be more precise its a subset of the linked PR with a "better" test plan.

.eslintrc.js
```diff
@@ -1,24 +1,29 @@
...
+  parser: "@typescript-eslint/parser",
   env: {
     browser: true,
     commonjs: true,
-    es6: true,
+    es2020: true,
```
bh2smith, line 6: Is this the cool new thing that includes two space indentation and no-semi colons?

bh2smith (approved):

.eslintrc.js
nlordell: No, this is needed for native `BigInt` support with the linter.

As far as I know, there is no "cool new thing" for indentation and semi-colons. While the large `.eslintrc` configuration file got removed, it was being ignore anyway and all of the configuration in there wasn't even being read. As a test I went back a couple commits and changed the `.eslintrc` file to be:
```
{
  asdofiadsi alfdoijalsidfh asdfli
```
And ran `yarn lint` and got no errors. (doing the same thing with the `.eslintrc.js` file  gives me syntax errors when I try to run the linter).

nlordell: I just checked, and `.eslintrc` without an extension has been deprecated as far back as 2018. Since both `.eslintrc` and `.eslintrc.js` existed when `.eslintrc` was [first added in 2019](https://github.com/gnosis/dex-contracts/tree/cb6b01ffb8a1cd31c0433a19ccfb27acc4edfd6a), the configuration in `.eslintrc`without an extension has always been ignored by our CI and never actually used.

It is possible that your editor prefers `.eslintrc` over `.eslintrc.js`, but that is not standard and not what our CI was doing anyway, so any rules in the `.eslintrc` without an extension have never actually been verified by Travis.

PR #704, opened 2020-04-30
[Dependencies] Yarn Upgrade

nlordell:
This PR just does a yarn upgrade and cleans up some of the dependencies in hopes to:
- Get it working with Node LTS (^12.0.0)
- Remove the "security warnings" banner in the repo.
- Pin dependencies that can't be updated (`@gnosis.pm/solidity-data-structures` because of breaking API changes and `openzeppelin-solidity` because of solc compiler versions)

We do this since dependabot doesn't automatically update transient dependencies.

### Test Plan

Check that CI still works.


e00E (approved):

fleupold (approved):

bh2smith (approved):
Amazeballs!

PR #705, opened 2020-04-30
Add type arithmetic to improve event type-safety

nlordell:
This PR just adds some utility types that improve type-safety on event types. They are currently not used, but I figured that it was worth including it in a separate PR since it is rather dense code that takes some time to get used to.

### Test Plan

Example of it in action:
```ts
const eventData: AnyEvent<BatchExchange> = {} as any
switch (eventData.event) {
case "Token":                              // ERR
  break
case "OrderPlacement":
  eventData.returnValues.buyToken = "asdf" // OK
  break
case "Withdraw":
  eventData.returnValues.buyToken = "asdf" // ERR
  break
}
```

Produces the folling `tsc` errors:
```
src/streamed/events.ts(50,8): error TS2678: Type '"Token"' is not comparable to type '"OrderPlacement" | "TokenListing" | "OrderCancellation" | "OrderDeletion" | "Trade" | "TradeReversion" |
 "SolutionSubmission" | "Deposit" | "WithdrawRequest" | "Withdraw"'.
src/streamed/events.ts(56,28): error TS2339: Property 'buyToken' does not exist on type '{ user: string; token: string; amount: string; 0: string; 1: string; 2: string; }'.
```

### DIFF
--- a/src/streamed/events.ts
+++ b/src/streamed/events.ts
@@ -16,14 +16,16 @@ type Event<
  * This type definition allows the TypeScript compiler to determine the type of
  * the `returnValues` property based on `event` property checks. For example:
  * ```
- * const eventData: AnyEvent<BatchExchange> // = ...
- * switch (eventData.event) {
- * case "TokenListing":
- *   x.returnValues.token = "..." // OK
+ * const eventData: AnyEvent<BatchExchange> = ...
+ * case "Token":                              // ERR: 2678: Type '"Token"' is not comparable to type
+ *                                            // '"OrderPlacement" | "TokenListing" | ...'
+ *   break
+ * case "OrderPlacement":
+ *   eventData.returnValues.buyToken = "asdf" // OK
  *   break
  * case "Withdraw":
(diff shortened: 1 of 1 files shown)


Velenir (approved):
We did the same at contract level [in dex-js](https://github.com/gnosis/dex-js/blob/master/src/contracts/types.ts#L27) :grin: 
I wonder if it makes sense to centralize the types somewhere

nlordell:
@Velenir I actually stole some of it from your implementation, but I didn't get a discriminated union for the `AllEvents` type from `type AllValues<T extends object> = T[keyof T]` But this could be because we have different generated contract types.

anxolin (approved):

bh2smith (approved):

fleupold (approved):
Very cool. Is this something that could be included upstream in the package that creates types for the contracts/web3 ?

nlordell:
> Very cool. Is this something that could be included upstream in the package that creates types for the contracts/web3 ?

Probably! I created #706 so we don't forget.

PR #707, opened 2020-04-30
Change re-exports

nlordell:
This PR changes how we re-export certain things in the `index.js` file:
- Artifacts are now re-exported as `${ContractName}Artifact`. **Note this is a breaking change**
- The generated contract typings are now re-exported from root as well. This makes using the `dex-contracts` module slightly easier with:
```
import { BatchExchange, BatchExchangeArtifact } from "@gnosis.pm/dex-contracts"

const { abi, networks } = BatchExchangeArtifact
const instance = new web3.eth.Contract(abi, networks[1].address) as BatchExchange
```

### Test Plan

This mostly changes our re-exports and doesn't have any real impact in the contracts code. This was just a pain-point when developing the event based orderbook where I had to dig into the package directory tree to find out where the types actually were.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,5 +1,8 @@
 import BN from "bn.js";
-import {BatchExchangeViewer} from "../build/types/BatchExchangeViewer";
+import { BatchExchangeViewer } from "../build/types/BatchExchangeViewer";
+
+export { BatchExchange } from "../build/types/BatchExchange";
+export { BatchExchangeViewer } from "../build/types/BatchExchangeViewer";
 export * from "./orderbook";
 export * from "./fraction";
 
@@ -54,7 +57,8 @@ export interface ContractArtifact {
   userdoc: any;
 }
 
-export declare const BatchExchange: ContractArtifact;
+export declare const BatchExchangeArtifact: ContractArtifact;
+export declare const BatchExchangeViewerArtifact: ContractArtifact;
 
 export interface Order {
   user: string;
--- a/src/index.js
+++ b/src/index.js
@@ -7,8 +7,8 @@
  */
 
 module.exports = {
-  BatchExchange: require("../build/contracts/BatchExchange.json"),
-  BatchExchangeViewer: require("../build/contracts/BatchExchangeViewer.json"),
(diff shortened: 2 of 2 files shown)


fleupold (approved):

e00E (approved):

anxolin:
Thanks for the heads up! 👍 

Why not `<ContractName>Contract`?
This is how we do it in dex-js:
https://github.com/gnosis/dex-js/blob/master/src/contracts/BatchExchangeContract.ts#L6

I wonder if some of the things we are doing in `dex-contract` would make sense to move them to `dex-js`. We'll be likely building small utils and tools that shouldn't be needing to increment version in the contracts, and probably many people won't need them when they require the NPM package. Also, probably they will keep adding dependencies in dex-contract and making the repo more complicated for a reader.

For us, `dex-js` is where we add all the "accesories", and it can be easily broken into multiple repositories in the future depending on need (or a mono), for example, you can make different packages wrapper in ethers.js, truffle, or web3, or evolping the order book logics won't affect the smart contracts.

nlordell:
@anxolin Sorry, just saw this now:

> Why not <ContractName>Contract?

This is specifically for the Truffle JSON artifact, and not an actual contract instance. It is a re-export of the data included in `build/contracts/BatchExchange.json`.

> I wonder if some of the things we are doing in dex-contract would make sense to move them to dex-js. 

I think, in the long run `dex-contracts` and `dex-js` could maybe merge into 1 repo that exports multiple packages. This would allow us to keep `dex-contracts` super light with just the contract ABI, typings, and encoding/decoding methods and no dependencies and move more utility funcitonality into a separate package.

Having a monorepo ensures that both packages are always kept up to date and breaking changes are handled immediately.

Also, I like your idea of having different "adapters" for ethers.js, truffle, web3, etc.

Let me know what you think.

anxolin:
> This is specifically for the Truffle JSON artifact, and not an actual contract instance. It is a re-export of the data included in build/contracts/BatchExchange.json.

Aha, ok!

> I think, in the long run dex-contracts and dex-js could maybe merge into 1 repo that exports multiple packages

Agree, I like it 👍

PR #730, opened 2020-05-05
[Events Orderbook] Implement orderbook update function

### LINKED ISSUE #718: [Events Orderbook] Polling fiber for keeping orderbook up to date

nlordell:
Implements a method that can be called to update the orderbook with the latest confirmed events.

This PR is an alternative to #724 where the update polling would be implemented by the caller instead.

Closes #718 

### Test Plan

Run the e2e test and see new confirmed events being applied:
```
fetching page 10000147-10006346
applying 1105 past events
fetching new events from 10006342-latest
applying 2 confirmed events until block 10006358
```

### DIFF
--- a/src/streamed/index.ts
+++ b/src/streamed/index.ts
@@ -47,12 +47,6 @@ export interface OrderbookOptions {
   blockConfirmations: number;
 
   /**
-   * Sets the polling interval used for querying latest events and updating the
-   * current account state.
-   */
-  pollInterval: number;
-
-  /**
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Nice, I think it makes sense to give this responsibility to the caller rather than secretly polling in the background. 
It might be nice to expose a way how to do this for convenience and reusability.

e00E (approved):

PR #736, opened 2020-05-06
[TypeScript] Lint TypeScipt code

nlordell:
Currently the typescript code is not being linted in the CI. This PR adds some basic recommended typescript linting rules and fixes to code so that the linting succeeds.

In order to fix an import lint on importing `truffle-assertions`, I added some typings for the module.

### Test Plan

CI - linting should succeed.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,33 +1,26 @@
 module.exports = {
+  root: true,
   parser: "@typescript-eslint/parser",
   env: {
     node: true,
     commonjs: true,
     es2020: true,
   },
-  extends: [
-    "eslint:recommended",
-    // "plugin:@typescript-eslint/eslint-recommended",
-    // "plugin:@typescript-eslint/recommended",
-  ],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/eslint-recommended", "plugin:@typescript-eslint/recommended"],
   parserOptions: {
-    ecmaFeatures: {
-      jsx: true,
-    },
     ecmaVersion: 2015,
     sourceType: "module",
   },
-  plugins: ["react", "@typescript-eslint"],
+  plugins: ["@typescript-eslint"],
   rules: {
     indent: ["error", 2],
     "linebreak-style": ["error", "unix"],
+    "no-console": "error",
     quotes: ["error", "double"],
     semi: ["error", "never"],
-    "prefer-const": ["error"],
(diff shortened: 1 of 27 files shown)


.eslintrc.js
```diff
@@ -1,34 +1,43 @@
...
     "linebreak-style": ["error", "unix"],
+    "no-console": "error",
     quotes: ["error", "double"],
     semi: ["error", "never"],
-    "prefer-const": ["error"],
-    "no-var": ["error"],
 
-    "no-unused-vars": "off",
-    "@typescript-eslint/no-unused-vars": ["error"],
+    //"prettier/prettier": "error",
```
e00E: Should this line be deleted?

e00E (approved):

fleupold (approved):
Very cool. It goes without saying that I didn't review all the ts changes, as I assume they were automated.

nlordell:
> Very cool. It goes without saying that I didn't review all the ts changes, as I assume they were automated.

I had to add type return types everywhere. The semi-colon stuff was automated.

I split this out into #737 and #738 which each contain some non-automatic/mechanical changes to appease the linter.

migration_conf.js
```diff
@@ -1,8 +1,6 @@
 const path = require("path")
 
-// eslint-disable-next-line no-undef
 const BUILD_DIR = path.join(__dirname, "build", "contracts")
-// eslint-disable-next-line no-undef
 const NETWORKS_FILE_PATH = path.join(__dirname, "networks.json")
```
bh2smith, line 4: Weird... why where these even here. Looks like they were perfectly defined..

src/balance_reader.ts
```diff
@@ -1,8 +1,8 @@
-import BN from "bn.js";
-import Web3 from "web3";
```
bh2smith, line 2: Awesome! I love no-semi

bh2smith (commented):
I would say that the semis could have been removed separately so I could actually detect what else was changed.

migration_conf.js
nlordell: Because the ESLint `env` was incorrectly setup for this file.

nlordell:
> I would say that the semis could have been removed separately so I could actually detect what else was changed.

The issue is that CI would fail as the formatting would be incorrect. I suggest filtering by `.json` and `.js` in the GitHub diff UI.

PR #737, opened 2020-05-06
[TypeScript] Add orderbook JSON types

nlordell:
This PR adds orderbook JSON types

This change was included as part of #736 but I think it deserves some special attention.

### Test Plan

Unit tests validate serialization and deserialization.

### DIFF
--- a/src/orderbook.ts
+++ b/src/orderbook.ts
@@ -1,4 +1,4 @@
-import { Fraction } from "./fraction";
+import { Fraction, FractionJson } from "./fraction";
 import BN from "bn.js";
 
 export class Offer {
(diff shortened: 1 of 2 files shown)


src/orderbook.ts
```diff
@@ -293,6 +301,14 @@ export class Orderbook {
   }
 }
 
+export interface OrderbookJson {
+  baseToken: string;
+  quoteToken: string;
+  remainingFractionAfterFee: Fraction | FractionJson;
+  asks: Record<string, Offer | OfferJson>;
+  bids: Record<string, Offer | OfferJson>;
```
e00E: Why allow this flexibility? If it comes from json then it should always be `OfferJson`, right?

e00E (approved):

src/orderbook.ts
nlordell: Its because `toJSON` method doesn't convert `Offer` to `OfferJson`.

nlordell: Hmm... I guess I can create 2 types here instead of just one. That might be a more correct approach.

nlordell: @e00E I split it into 2 types now. Let me know if you think this is better.

PR #739, opened 2020-05-06
[Events Orderbook] Use token array instead of map

nlordell:
This PR refactors the auction state to use a token array instead of a map so that it is trivial to check if a token listing was missed.

### Test Plan

New unit tests added


e00E (approved):

fleupold (approved):
Thanks for splitting. This change by itself is very understandable

bh2smith (commented):
Clean, manageable and easy to review. Very nice.

PR #741, opened 2020-05-06
[TypeScript] Format TS files

nlordell:
This PR adds `prettier` formatting to TS files as well. Note that we no longer check JS and TS formatting with `yarn pretty-check` since `eslint` already checks formatting with the `prettier` plugin. The `.sol` formatting check was moved to a the solidity linting script.

Note, all code changes were done automatically by running `yarn prettier`

### Test Plan

CI should succeed without formatting issues.


.eslintrc.js
```diff
@@ -6,14 +6,20 @@ module.exports = {
...
   parserOptions: {
-    ecmaVersion: 2015,
+    ecmaVersion: 2020,
     sourceType: "module",
   },
-  plugins: ["@typescript-eslint"],
+  plugins: ["@typescript-eslint", "prettier"],
   rules: {
-    indent: ["error", 2],
+    indent: ["error", 2, { SwitchCase: 1 }],
```
bh2smith: So indentation is 1 for switch statements? Do we have any of these so I can see what it looks like?

test/models/streamed/index.spec.ts
```diff
@@ -5,21 +5,16 @@ import { StreamedOrderbook } from "../../../src/streamed"
...
   describe("init", () => {
-    it("should successfully apply all events and match on-chain orderbook", async function() {
-      const {
-        ETHEREUM_NODE_URL,
-        INFURA_PROJECT_ID,
-        ORDERBOOK_END_BLOCK,
-        TEST_STREAMED_ORDERBOOK_E2E,
-      } = process.env
+    it("should successfully apply all events and match on-chain orderbook", async function () {
+      const { ETHEREUM_NODE_URL, INFURA_PROJECT_ID, ORDERBOOK_END_BLOCK, TEST_STREAMED_ORDERBOOK_E2E } = process.env
```
bh2smith: This is why we are going to go with #742

test/models/streamed/index.spec.ts
```diff
@@ -5,21 +5,16 @@ import { StreamedOrderbook } from "../../../src/streamed"
...
       if (!TEST_STREAMED_ORDERBOOK_E2E) {
         this.skip()
         return
       }
 
       assert.isDefined(
         ETHEREUM_NODE_URL || INFURA_PROJECT_ID,
-        "ETHEREUM_NODE_URL or INFURA_PROJECT_ID environment variable is required",
+        "ETHEREUM_NODE_URL or INFURA_PROJECT_ID environment variable is required"
```
bh2smith: Ya... we gotta have this trailing comma.

test/resources/array-shims.ts
```diff
@@ -10,13 +10,7 @@
...
   // implementation taken from
   // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#Alternative
-  return depth > 0
-    ? arr.reduce(
-      (acc, val) =>
-        acc.concat(Array.isArray(val) ? flat(val, depth - 1) : val),
-      []
-    )
-    : arr.slice()
+  return depth > 0 ? arr.reduce((acc, val) => acc.concat(Array.isArray(val) ? flat(val, depth - 1) : val), []) : arr.slice()
```
bh2smith: Totally!

.travis.yml
```diff
@@ -11,8 +11,6 @@ before_script:
   - yarn run ganache > /dev/null &
 script:
   - yarn lint
-  - yarn solhint
```
bh2smith, line 14: I know you mentioned something about this in the description, but I don't see a replacement for this here. Is it already somewhere else?

bh2smith (approved):

.travis.yml
nlordell: https://github.com/gnosis/dex-contracts/pull/741/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R24-R30

.eslintrc.js
nlordell: Ah, thanks for pointing this out - it should actually be removed as we want `prettier` plugin to enforce indentation so that we don't get issues where the auto-code formatter wants code indented one way and ESLint wants it another way.

Also, the `1` is indentation level. So if indentation is 2 spaces, then `SwitchCase` of 0 is:
```
switch (foo) {
case "bar":
  console.log("bar")
  break
}
```

`SwitchCase` of 1 is:
```
switch (foo) {
  case "bar":
    console.log("bar")
    break
}
```

`SwitchCase` of 2 is:
```
switch (foo) {
    case "bar":
      console.log("bar")
      break
}
```

It was set up this way because `prettier` does not allow configuring this behaviour and always indents like the `SwitchCase` of 1

fleupold (approved):

PR #742, opened 2020-05-06
[Contraversial] Propose new formatting

nlordell:
This PR proposes new code auto-formatting rules for the TS code. Comments inline.

Overall, this PR falls back to using more of the default values instead of customizing the formatting too much.

All code changes were done with `npx prettier --write ...`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,6 @@ module.exports = {
   },
   plugins: ["@typescript-eslint", "prettier"],
   rules: {
-    "linebreak-style": ["error", "unix"],
     "no-console": "error",
 
     "@typescript-eslint/camelcase": "off",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,4 +41,4 @@ module.exports = {
     },
   ],
   ignorePatterns: ["build/", "coverage/", "node_modules/"],
-};
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,11 +19,8 @@ module.exports = {
   },
   plugins: ["@typescript-eslint", "prettier"],
   rules: {
-    indent: ["error", 2, { SwitchCase: 1 }],
     "linebreak-style": ["error", "unix"],
     "no-console": "error",
-    quotes: ["error", "double"],
-    semi: ["error", "never"],
 
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/no-use-before-define": "off",
@@ -44,4 +41,4 @@ module.exports = {
     },
   ],
   ignorePatterns: ["build/", "coverage/", "node_modules/"],
-}
+};
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -12,5 +12,13 @@ module.exports = {
         printWidth: 129,
       },
     },
+    {
+      files: "*.js",
+      options: {
+        printWidth: 129,
+        semi: false,
+        trailingComma: "es5",
+      },
+    },
   ],
-};
+}
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,17 +1,16 @@
 module.exports = {
+  // NOTE: This is actually the default value but it is being overwritten by
+  // the solidity plugin somehow.
   bracketSpacing: true,
-  printWidth: 129,
-  semi: false,
-  trailingComma: "es5",
+  trailingComma: "all",
 
   overrides: [
     {
       files: "*.sol",
       options: {
         bracketSpacing: false,
-        explicitTypes: "always",
-        tabWidth: 4,
+        printWidth: 129,
       },
     },
   ],
-}
+};
--- a/.solcover.js
+++ b/.solcover.js
@@ -10,4 +10,4 @@ module.exports = {
     grep: "@skip-on-coverage", // Find everything with this tag
     invert: true, // Run the grep's inverse set.
   },
-}
+};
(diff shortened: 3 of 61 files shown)


.prettierrc.js
```diff
@@ -1,17 +1,16 @@
...
-  trailingComma: "es5",
+  trailingComma: "all",
 
   overrides: [
     {
       files: "*.sol",
       options: {
         bracketSpacing: false,
-        explicitTypes: "always",
-        tabWidth: 4,
```
nlordell, line 13: These two are the plugin defaults and including them here is redundant.

.prettierrc.js
```diff
@@ -1,17 +1,16 @@
 module.exports = {
+  // NOTE: This is actually the default value but it is being overwritten by
+  // the solidity plugin somehow.
   bracketSpacing: true,
-  printWidth: 129,
-  semi: false,
-  trailingComma: "es5",
+  trailingComma: "all",
```
nlordell, line 5: We already enable trailing commas in most places (specifically all places supported by ES5, so objects, imports, arrays) and just not in function arguments:
```
foo(
  a,
  b
)
```
becomes:
```
foo(
  a,
  b,
)
```

This is beneficial for the same reason the other currently enabled trailing commas are beneficial. Specifically, the main argument for them is that they improve diffs as adding new parameters only changes one line instead of two.

I imagine one of the reasons `all` is not the default is that it requires [NodeJS >8 to support this](https://prettier.io/docs/en/options.html#trailing-commas). This is a non-issue for us since TS should generate compatible code regardless of formatting.

.prettierrc.js
```diff
@@ -1,17 +1,16 @@
 module.exports = {
+  // NOTE: This is actually the default value but it is being overwritten by
+  // the solidity plugin somehow.
   bracketSpacing: true,
-  printWidth: 129,
```
nlordell, line 3: Remove the current print width and set it to default of 80. `prettier` has a section on its rational for using 80 character width [here](https://prettier.io/docs/en/options.html#print-width) and [here](https://prettier.io/docs/en/rationale.html#print-width).

.prettierrc.js
```diff
@@ -1,17 +1,16 @@
...
-  semi: false,
```
nlordell, line 4: Not using semicolons in JavaScript/TypeScript can lead to ambiguous code. The `prettier` documentation [has a good example outlining this](https://prettier.io/docs/en/rationale.html#semicolons).

Additionally, semicolons are more common in TS and the more standard style, this includes:
- [Google TypeScript style](https://github.com/google/gts)
- [Airbnb style](https://github.com/airbnb/javascript#semicolons)
- [Microsoft TypeScript compiler](https://github.com/microsoft/TypeScript/blob/master/src/tsc/tsc.ts)

All in all, I think it is a more commonly used style and doesn't suffer from parsing ambiguity issues. After all, in TS whitespace is not significant.

.eslintrc.js
```diff
@@ -19,11 +19,8 @@ module.exports = {
   },
   plugins: ["@typescript-eslint", "prettier"],
   rules: {
-    indent: ["error", 2, { SwitchCase: 1 }],
     "linebreak-style": ["error", "unix"],
     "no-console": "error",
-    quotes: ["error", "double"],
-    semi: ["error", "never"],
```
nlordell: These rules can be safely removed as they are enforced by the `prettier` plugin.

.prettierrc.js
bh2smith: I like this.

.prettierrc.js
bh2smith: Will this affect the smart contract of just the JS? I am ok with the JS, but not the smart contract.

bh2smith: The second link says something about prettier sometimes ignoring lines. I was not aware of this without having to explicitly declare a line to be ignored which I find ugly.

.prettierrc.js
bh2smith: ~~Fine. As long as we pick one and stay there. I suppose all my contributions to this project have already been overwritten in the annotations anyway so whats another 80% of the entire code base.~~

Ya, I only noticed later.

.eslintrc.js
```diff
@@ -19,11 +19,8 @@ module.exports = {
   },
   plugins: ["@typescript-eslint", "prettier"],
```
bh2smith: Wasn't the prettier removed in the last PR?

bh2smith (approved):
To me it looks like a few lint configurations followed by adding all the semi colons back. Love the trailing comma btw!

.prettierrc.js
nlordell: From what I understand it has specific logic for imports and unit tests, but for other things you still need to manually ignore a line.

nlordell: Also this change only affects TS moving forward, JS and Sol should remain unchanged.

.prettierrc.js
nlordell: Again, this only affects TS, JS remains as is. I think the only changed JS files in the diff are the prettier and ESLint config files

.eslintrc.js
nlordell: It was added last PR and the `prettier --check` was removed because `eslint` verifies it internally

nlordell:
> To me it looks like a few lint configurations followed by adding all the semi colons back. 

Yes, the idea was the previous PR would just setup the code formatter for TS with the current JS settings, and this PR would be where we can decide what we want the TS settings to be.

fleupold (approved):
I'm on board with the proposed changes. No strong feelings on the semicolons.

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FxSM46ernAUN3y%2Fgiphy-downsized-medium.gif&token=whatever_you_say" data-gitmeme-token="whatever_you_say"></a>

PR #749, opened 2020-05-11
[Easy] Simplify streamed orderbook logging options

nlordell:
This PR just simplifies the `OrderbookOptions` to accept a debug function instead of a full `logger` type as it is only really useful for logging internal progress (like which page it is on when building the streamed orderbook).

There used to be a call to `logger.error` as well for unrecoverable errors, but that doesn't have much value as the error is propagated as an exception and is usually logged by the caller.

### Test Plan

CI, thanks to TypeScript type safety!

### DIFF
--- a/src/streamed/index.ts
+++ b/src/streamed/index.ts
@@ -64,14 +64,10 @@ export interface OrderbookOptions {
   strict: boolean;
 
   /**
-   * Set the logger to be used by the streamed orderbook module.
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

bh2smith (approved):

e00E (approved):

fleupold (approved):
🍬

PR #760, opened 2020-05-13
[Easy] Use `import type` where appropriate

nlordell:
This PR changes the code to use `import type` where appropriate. This makes it so the generated `.js` code doesn't `require(...)` the package. This is useful when using a package just for type definitions.

### Test Plan

CI - no code change. Note the compiler doesn't let you use `import type` if the imported piece is actually used.

### DIFF
--- a/src/balance_reader.ts
+++ b/src/balance_reader.ts
@@ -1,6 +1,6 @@
 import BN from "bn.js";
-import Web3 from "web3";
-import { BatchExchangeInstance } from "../types/truffle-typings";
+import type Web3 from "web3";
+import type { BatchExchangeInstance } from "../types/truffle-typings";
 
 const WORD_DATA_LENGTH = 64;
 
--- a/src/onchain_reading.ts
+++ b/src/onchain_reading.ts
@@ -4,9 +4,8 @@ import {
   Order,
   IndexedOrder,
 } from "./encoding";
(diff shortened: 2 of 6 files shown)


fleupold (approved):
🦄

bh2smith (approved):
Is this kind of import strictly enforced, or is it just part of the style guide? Is there a reason we didn't use it on the `BN` import? Is `BN` not a "type"? What if a file exports more than just types?

nlordell:
> Is this kind of import strictly enforced

Yes, you can only use it for importing types, if you try to actually use code imported this way (like calling a constrcutor or a function directly) the compiler will error.

> Is there a reason we didn't use it on the BN import?

In the specific situation where use used `import type BN from "bn.js"` we only needed to import the BN type so that the compiler could ensure that we were calling the correct methods on it, but that specific module never actually creates a `BN` and just passes them around. This way the generated JS code no longer has a `require("bn.js")` statement (since it wasn't actually needed in that module anyway!).

> What if a file exports more than just types?

That is the purpose of `import type` - we want to import only type information from a file, but not actually call any of its code directly.

bh2smith:
> That is the purpose of import type - we want to import only type information from a file, but not actually call any of its code directly.

This is the most interesting thing I learned about this today. Thanks!

PR #762, opened 2020-05-13
[Easy] Remove minimist dependency

### LINKED ISSUE #715: [Dependencies] Remove `minimist` dependency.
We currently use `yargs` for command line parsing almost everywhere except for `truffle-config.js` where we use the `minimist` package. We should probably just use `yargs` everywhere because of the pirate theme, and it is the most commonly used command line parser in our package.

nlordell:
We were using `minimist` in only one place for parsing command line arguments and `yargs` everywhere else. This PR removes the minimist dependency and replaces its single use with `yargs` cause pirates are cool.

Closes #715

### Test Plan

Run:
```
npx truffle test --gas --grep @skip-on-coverage
```

And see only `@skip-on-coverage` tests being run and an `eth-gas-reporter` summary at the end.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,6 @@
     "eth-gas-reporter": "^0.2.17",
     "ganache-cli": "^6.9.1",
     "husky": "^4.2.5",
-    "minimist": "^1.2.5",
     "mocha": "^7.1.1",
     "node-fetch": "^2.6.0",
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Nice find! Really great! Pirates Rule 🏴‍☠️

e00E (approved):

PR #764, opened 2020-05-13
[Trivial] Remove debug code that snuck in

nlordell:
This PR removes some debug code that snuck in but shouldn't be there: the streamed orderbook test was effectively only comparing the first 10 orders :see_no_evil: 

### Test Plan

The streamed orderbook test should still succeed!
```
yarn test-streamed-orderbook
```


fleupold (approved):
🙈 🙉 🙊

PR #765, opened 2020-05-13
[WIP] Verify streamed orderbook

nlordell:
This PR adds a new script a streamed orderbook and periodically checks that it matches the on-chain version of the orderbook.

This will hopefully give us some confidence about the new streamed orderbook implementation.

### Test Plan

It is the test!
```
npx truffle exec --network mainnet scripts/verify_streamed_orderbook.js
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:lts-alpine
+
+RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
+
+COPY yarn.lock package.json ./
+RUN yarn install --frozen-lockfile && yarn cache clean
+
+COPY . .
+RUN yarn prepack
+
+ENTRYPOINT ["bash", "-c"]
+CMD ["npx", "truffle", "exec", "scripts/verify_streamed_orderbook.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,14 @@ FROM node:lts-alpine
 
 RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN apk add --no-cache --virtual build-dependencies bash git python make g++ ca-certificates
+
+COPY yarn.lock package.json ./
+RUN yarn install --frozen-lockfile && yarn cache clean
+
+COPY . .
+RUN yarn prepack
+
+ENTRYPOINT ["bash", "-c"]
+CMD ["npx", "truffle", "exec", "scripts/verify_streamed_orderbook.js"]
```
bh2smith: No newline at end of file!?

bh2smith (approved):
Really cool. The terminology used reminds me of a pure mathematical construct I studied extensively in grad school. That is, ["Fibre Bundles"](https://en.wikipedia.org/wiki/Fiber_bundle) or more exotically, [Vector](https://en.wikipedia.org/wiki/Vector_bundle) and [Principal](https://en.wikipedia.org/wiki/Principal_bundle) bundles

Dockerfile
nlordell: There is one... When there isn't there is a little symbol: 
image

nlordell:
> Fiber bundles

A "fiber" is [lightweight thread of execution](https://en.wikipedia.org/wiki/Fiber_(computer_science)) made up of promises. This is different than a background thread which would require a `Worker` and is more heavy-weight.

Dockerfile
bh2smith: Nice symbol!

PR #772, opened 2020-05-14
Added order encoding methods

nlordell:
This PR implements order encoding in `dex-contracts`. This can be used in `dex-price-estimator` once released to an npm package.

### Test Plan

New unit tests.

### DIFF
--- a/src/encoding.ts
+++ b/src/encoding.ts
@@ -8,7 +8,7 @@
 import BN from "bn.js";
 import assert from "assert";
 
-export interface Order<T = string> {
+export interface Order<T = BN> {
   user: string;
   sellTokenBalance: T;
   buyToken: number;
@@ -20,7 +20,7 @@ export interface Order<T = string> {
   remainingAmount: T;
 }
 
-export interface IndexedOrder<T> extends Order<T> {
+export interface IndexedOrder<T = BN> extends Order<T> {
   orderId: number;
 }
 
@@ -60,7 +60,7 @@ function decodeIndexedOrder(bytes: OrderBuffer): IndexedOrder<BN> {
 }
 
 function decodeOrdersInternal<T>(
-  bytes: string,
+  bytes: string | null,
   decodeFunction: (x: OrderBuffer) => T,
   width: number,
 ): T[] {
@@ -82,7 +82,7 @@ function decodeOrdersInternal<T>(
  * decode the result of `BatchExchange.getEncodedUserOrders` and
  * `BatchExchange.getEncodedOrders`.
  */
-export function decodeOrders(bytes: string): Order<BN>[] {
+export function decodeOrders(bytes: string | null): Order<BN>[] {
   return decodeOrdersInternal<Order<BN>>(bytes, decodeOrder, 112);
 }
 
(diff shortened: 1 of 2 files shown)


e00E (approved):

test/models/encoding.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+          validFrom: 4,
+          validUntil: 5,
+          priceNumerator: new BN(1337),
+          priceDenominator: new BN(0x12345678),
+          remainingAmount: new BN(42),
+        },
+      ];
+      const encoded = encodeOrders(orders);
+      const decoded = decodeOrders(hex(encoded));
+      expect(json(decoded)).to.deep.equal(json(orders));
```
fleupold, line 58: do we need to compare the json here or could we just deep compare the objects directly?

fleupold (approved):
Nice, awesome unit tests!

test/models/encoding.spec.ts
nlordell: We need to normalize the objects in some way because different methods for constructing BN lead to different internal representations of the same number. Specifically, constructing from a number directly creates a BN with a single digit, while constructing from a hex string adds leading 0 digits:

```
> assert.deepEqual(new BN(0x2a), new BN("2a", 16))
Uncaught AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

BN {
  length: 1,
  negative: 0,
  red: null,
  words: [
    42
  ]
}

should loosely deep-equal

BN {
  length: 1,
  negative: 0,
  red: null,
  words: [
    42,
    0
  ]
}
```

Since `BN` provides a `toJSON` implementation, it gets properly converted to a normalized implementation when comparing the JSON objects:
```
> assert.deepEqual(JSON.stringify(new BN(0x2a)), JSON.stringify(new BN("2a", 16)))
undefined
```

nlordell: This is why I'm all for dropping `BN` for `BigInt` ;)

fleupold: > This is why I'm all for dropping BN for BigInt ;)

+1 Maybe we can create a cleanup task for that?

src/encoding.ts
```diff
@@ -82,7 +82,7 @@ function decodeOrdersInternal<T>(
  * decode the result of `BatchExchange.getEncodedUserOrders` and
  * `BatchExchange.getEncodedOrders`.
  */
-export function decodeOrders(bytes: string): Order<BN>[] {
+export function decodeOrders(bytes: string | null): Order<BN>[] {
```
bh2smith, line 85: Nice!

test/models/encoding.spec.ts
bh2smith: Thats gonna be a bigun'

bh2smith:
Although this is already merged, this seems like an appropriate place to raise my struggles with running `yarn run test-ts`

~~I keep getting some error with `web3` Is there something I need to do first?  Like `build`?~~

Yes, `yarn build` did the trick.

Also... It looks like the event orderbook test is "pending"

```
  76 passing (291ms)
  1 pending

✨  Done in 6.13s.
```

How do we know if it ever completed?

nlordell:
> Yes, yarn build did the trick.

The actual missing bit was the TypeScript bindings to the SCs. I think the trick with that is `yarn ts-contracts`.

> Also... It looks like the event orderbook test is "pending"

For mocha, a "pending" test is one that is skipped. This is because we have a long running test that builds a complete event-based orderbook and compares it to the EVM queried orderbook. You can run it with `yarn test-streamed-orderbook`. The test is in `test/models/streamed/index.spec.ts`.

PR #778, opened 2020-05-18
Test multiply touched orders in a single trade

### LINKED ISSUE #777: [Testing] Solution touches same order multiple times.
It was recently pointed out that our smart contract does not impose a uniqueness constraint on the orders touched in a given solution. This implies that the following scenario could occur:

**Orders**
- user_1 sell x_1 `tokenA` for y_1  `tokenB`
- user_2 sell x_2 `tokenA` for y_2 `tokenB`
- user_3 sell z `tokenB` for y `tokenA`

**Solution:** (without consideration for fees)

```
owners = [user_1, user_2, user_3, user_3]
buyAmounts = [y1, y2, z /2, z/2]
```

Or even simpler, in a direct trade with only two users, a single trade could be split up with a higher multiplicity for no apparent reason. A common term used to describe this in algebraic geometry would be _ramification index_

Not that there was ever an error with this, but this issue is to add a unit (or regression) test that everything turns out as expected with solutions with ramification index greater than 1 are proposed.

bh2smith:
As we have recently learned, it is possible to "touch" the same order twice in a single solutions submission. This is elaborated on in the related issue:

Closes #777

Tagging @alfetopito and @anxolin who brought this up. I think this new discovery may also make it difficult to query for "finalized trades"

_Solutions with non-trivial touched order ramification index_

Test Plan: Unit tests.

### DIFF
--- a/test/regression_tests.js
+++ b/test/regression_tests.js
@@ -98,17 +98,19 @@ contract("BatchExchange", async (accounts) => {
 
       const solution = solutionSubmissionParams(basicTrade.solutions[0], accounts, orderIds)
 
+      // Modifying owners so that last owner appears twice.
+      // That is, owners = [owner1, owner2, ... ownerN, ownerN]
+      solution.owners.push(solution.owners.slice(-1)[0])
+      // Modifying touched orders so that last orderID appears twice.
+      solution.touchedorderIds.push(solution.touchedorderIds.slice(-1)[0])
       // Replace the last element of the basic trade volume with two coppies of half of it.
       const lastVolume = solution.volumes.slice(-1)[0]
-      const ramifiedVolumes = solution.volumes.slice(-1).concat([lastVolume.divn(2), lastVolume.divn(2)])
+      const ramifiedVolumes = solution.volumes.slice(0, -1).concat([lastVolume.divn(2), lastVolume.divn(2)])
 
       const ramifiedSolution = {
         objectiveValue: solution.objectiveValue,
-        // Modifying owners so that last owner appears twice.
-        // That is, owners = [owner1, owner2, ... ownerN, ownerN]
-        owners: solution.owners.push(solution.owners.slice(-1)[0]),
-        // Modifying touched orders so that last orderID appears twice.
-        touchedorderIds: solution.touchedorderIds.push(solution.owners.slice(-1)[0]),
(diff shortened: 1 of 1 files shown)


test/regression_tests.js
```diff
@@ -87,5 +87,44 @@ contract("BatchExchange", async (accounts) => {
...
+      const lastVolume = solution.volumes.slice(-1)[0]
+      const ramifiedVolumes = solution.volumes.slice(-1).concat([lastVolume.divn(2), lastVolume.divn(2)])
+
+      const ramifiedSolution = {
+        objectiveValue: solution.objectiveValue,
+        // Modifying owners so that last owner appears twice.
+        // That is, owners = [owner1, owner2, ... ownerN, ownerN]
+        owners: solution.owners.push(solution.owners.slice(-1)[0]),
+        // Modifying touched orders so that last orderID appears twice.
+        touchedorderIds: solution.touchedorderIds.push(solution.owners.slice(-1)[0]),
```
bh2smith:
```suggestion
        touchedorderIds: solution.touchedorderIds.push(solution.touchedorderIds.slice(-1)[0]),
```

anxolin (commented):
If I understand the issue, you are saying that:

A solution of a batch where Bob sells 100 DAI to Alice for OWL will be solvable with 2 trades
`Bob-Alice 100DAI + Alice-Bob 100 OWL` (let me overlook fees for the example)

But although silly, and our solver won't do it, the contract will accept:
`Bob-Alice 100DAI + Alice-Bob 50 OWL + Alice-Bob 50 OWL` 

Is that it?

If that's the case, I think my strategy still works, although I might have to review the PK of the table.

Imagine even an edge case where for a single batch we have 3 solutions, first one touches the order 1 time, the other one 2 times.

Trades: solution_1_trade,  solution_2_trade_1, solution_2_trade_2,  solution_3_trade_1, solution_3_trade_2
Reverts: solution_1_trade,  solution_2_trade_1, solution_2_trade_2

If we sort stuff by block and event index, and assign a cardinal number, we can relate the reverts to its trade

fedgiac (approved):
I'm not familiar with the details of the default testing utils that are used here, like `makeDeposits` or `solutionSubmissionParams`, but the logic of the test looks good to me.

PR #782, opened 2020-05-20
[Easy] Prep new npm version

bh2smith:
In a recent PR, it was forgotten to export the new amounts utility functions in the `src/index.ts` This is done here in preparation for a new npm version.

**Test Plan**

- Checkout this branch and run `yarn link` (note you may have to unlink first)
- Change in to, for example, `dex-liquidity` project directory and checkout branch `184/move_utils`
- run `yarn link "@gnosis.pm/dex-contracts"`
- run something like, `truffle test`

### DIFF
--- a/src/amounts.ts
+++ b/src/amounts.ts
@@ -1,6 +1,6 @@
 import BN from "bn.js";
 import { assert } from "chai";
-import { Fraction } from "../fraction";
+import { Fraction } from "./fraction";
 
 const MAX128 = new BN(2).pow(new BN(128)).subn(1);
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@
  * @packageDocumentation
  */
 
+export * from "./amounts";
 export * from "./balance_reader";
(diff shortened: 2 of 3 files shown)


fedgiac (approved):
I wonder if `src/index.ts` could/should be generated automatically from the content of `src/`.
A cursory search seems to suggest that you can't just export everything in a folder using TS. 
Do we ever import directly from this file in this repo? If not, `src/index.ts` could simply be generated every time we publish a new version of the package.

bh2smith:
> I wonder if src/index.ts could/should be generated automatically from the content of src/.
A cursory search seems to suggest that you can't just export everything in a folder using TS.
Do we ever import directly from this file in this repo? If not, src/index.ts could simply be generated every time we publish a new version of the package.

I believe there are scenarios where we would not want to make certain methods public. For example, I would actually prefer if `toERC20Units` was not public.. I guess it isn't because it is not "experted" explicitly in the file containing its declaration. Ok then, yea I now wonder the same thing.

PR #784, opened 2020-05-20
use JS assert and bump version

bh2smith:
importing chai as a main project dependency was overkill for such simple assertions in `amounts.ts` Hopefully this works for those exported functions in integrated projects.

### DIFF
--- a/test/models/amounts.spec.ts
+++ b/test/models/amounts.spec.ts
@@ -5,7 +5,7 @@ import {
 } from "../../src/amounts";
 import { Fraction } from "../../src/fraction";
 import BN from "bn.js";
-import { assert } from "chai";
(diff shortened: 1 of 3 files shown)


test/models/amounts.spec.ts
```diff
@@ -5,7 +5,7 @@ import {
 } from "../../src/amounts";
 import { Fraction } from "../../src/fraction";
 import BN from "bn.js";
-import { assert } from "chai";
+import assert from "assert";
```
nlordell: Unit tests should continue to use `chai` since it is a more powerful assertion library. Just the code in `src/` shouldn't.

nlordell (approved):

PR #797, opened 2020-05-26
[JS->TS] Migrate regression_tests.js

bh2smith:
In partial fulfillment of #330. Note that one test has been moved into a more appropriate place and one of the lint ignores was acting strange on my side so I moved them to be more directly in line with there intended purpose.

**Test Plan:** Unit tests and linter.


test/utilities.ts
```diff
@@ -53,9 +53,10 @@ export async function closeAuction(
 }
 
 export async function sendTxAndGetReturnValue<T>(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   method: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     sendTransaction: (...args: any[]) => Promise<string>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nlordell: nit: This shouldn't be needed anymore if you merge with master:
https://github.com/gnosis/dex-contracts/blob/master/test/utilities.ts#L55-L67

nlordell (approved):
Awesome! Just one minor nit with the disabled lints - they should already be fixed in master.

test/utilities.ts
bh2smith: Awesome thanks. Was that recent? I just branch off master this morning (or so I thought)

nlordell: It was from this morning.

PR #801, opened 2020-05-27
Upgrade to TypeChain v2

nlordell:
This PR updates the TypeChain targets to the v2 versions. This will hopefully keep us getting TypeChain improvements. For example, the `Linkable` type and cast is no longer needed :).

### Test Plan

CI, this should only affect generated typings and contains no real code modifications (besides adaptations to use new types).


e00E (approved):

fleupold (approved):
Nice, less custom types are great!

test/BatchExchange.spec.ts
```diff
@@ -72,14 +71,8 @@ contract("BatchExchange", async (accounts) => {
...
-    await (BatchExchange as Linkable<BatchExchangeInstance>).link(
-      "IdToAddressBiMap",
-      lib1.address,
-    );
-    await (BatchExchange as Linkable<BatchExchangeInstance>).link(
-      "IterableAppendOnlySet",
-      lib2.address,
-    );
+    await BatchExchange.link("IdToAddressBiMap", lib1.address);
+    await BatchExchange.link("IterableAppendOnlySet", lib2.address);
```
bh2smith, line 75: Cool! I am happy about this!

test/BatchExchange.spec.ts
```diff
@@ -166,9 +159,7 @@ contract("BatchExchange", async (accounts) => {
       );
     });
     it("Burns 10 OWL when adding token", async () => {
-      const TokenOWLProxy = artifacts.require(
-        "../node_modules/@gnosis.pm/owl-token/build/contracts/TokenOWLProxy",
-      );
+      const TokenOWLProxy = artifacts.require("TokenOWLProxy");
```
bh2smith, line 162: Weird, I hadn't even noticed this was a thing. What was making the OWL token proxy so special before?

types/truffle-typings/index.d.ts
```diff
@@ -2,10 +2,14 @@
...
 
-export interface Linkable<T> extends Truffle.Contract<T> {
-  link(lib: string, address: string): Promise<void>;
+declare global {
+  namespace Truffle {
+    interface Artifacts {
+      require(name: string): Truffle.Contract<unknown>;
+    }
+  }
 }
```
bh2smith: I guess this is going to conflict with my deposit migration #802. Should be easy enough to deal with the conflict on either side.

bh2smith (approved):
Awesome stuff!

test/BatchExchange.spec.ts
nlordell: Nothing, Truffle works with both methods. The main difference is that when you do it this way it works better with the generated types.

types/truffle-typings/index.d.ts
nlordell: Trivial to merge.

PR #802, opened 2020-05-27
[JS->TS] MIgrate scripts/deposit.js

bh2smith:
This is the first in a series of script migrations from JS to TS. This PR lays out the ground work for the entire migration by first including a new truffle-type for `callbacks`. Note also the temporary adjustment to `scripts/.eslintrc.js` which allows console log in JS files, but not in TS. INstead we are using `typescript-logging` which first appeared in #798

**Test Plan:** One unfortunate drawback of this Truffle interactions with Typescript is that we have to call the file from the build folder.

```sh
yarn && yarn build
npx truffle exec build/common/scripts/deposit.js --network rinkeby --accountId 0 --tokenId 1 --amount 0.01
```
Sit back and watch the magic! Should expect to see something like:

```
2020-05-27 12:50:46,052 INFO [deposit] Using account 0x627306090abaB3A6e1400e9345bC60c78a8BEf57
2020-05-27 12:50:57,353 INFO [deposit] Submitting approval for amount 100000000000000000
[HDWalletProvider] Using nonce:  25841
[HDWalletProvider] Using nonce:  25842
2020-05-27 12:51:29,292 INFO [deposit] Deposit successful! Can be traded as of batch 5301922
```

Partial fulfillment of #330


fleupold (approved):
Nice and clean. It would be cool if we could create an alias such as `yarn script <path>.ts ...args` which gets translated into `yarn truffle exec /build/common/<path>.js ...args` (and potentially compiles before executing it)

bh2smith:
> It would be cool if we could create an alias such as yarn script <path>.ts ...args which gets translated into yarn truffle exec /build/common/<path>.js ...args (and potentially compiles before executing it)

Yes this is a cool idea. I believe this is something @nlordell is actively working on at the moment. Perhaps I will not interfere there and see how that rolls out!

PR #806, opened 2020-05-27
[JS->TS] Migrate request_withdraw.js

bh2smith:
Very few changes made (added additional log "Using account ...". Partially fulfills #330

**Test Plan:**

```sh
npx truffle exec build/common/scripts/request_withdraw.js --network rinkeby --accountId 0 --tokenId 1 --amount 1
```

Should see something like 

```
Using network 'rinkeby'.

[HDWalletProvider] Using nonce:  25844
Withdraw request successful. Will be claimable in batch 5301964
```


scripts/request_withdraw.ts
```diff
@@ -0,0 +1,46 @@
...
+    const withdrawer = accounts[argv.accountId];
+    log.info(`Using account ${withdrawer}`);
+
+    const token_address = await instance.tokenIdToAddressMap(argv.tokenId);
+    if (token_address === ZERO_ADDRESS) {
+      callback(`Error: No token registered at index ${argv.tokenId}`);
+    }
+
+    await instance.requestWithdraw(token_address, amount, { from: withdrawer });
+    const claimable_at = (
```
nlordell: Not sure how you feel about array destructuring but you should be able to write 
```ts
const [, claimable_at] = await instance.getPendingWithdraw(...);
```

It’s perfectly fine how it is now, but sometimes using array destructuring can look a little nicer.

nlordell (approved):
Awesome sauce

scripts/request_withdraw.ts
bh2smith: Will fix ~~in the morning!~~ right now.

PR #808, opened 2020-05-27
[JS->TS] Migrate close_auction.js

bh2smith:
This is arguably unnecessary since this script can only be used on a local blockchain (or something with `evm_increaseTime`. It was originally intended for bash scripting our e2e tests in `dex-services`. Either way, the migration is here and there are very few changes to the code. In partial fulfillment of #330

**Test Plan / Script Usage**

```sh
npx truffle exec build/common/scripts/close_auction.js
```


nlordell (approved):
Owning the TS migration

PR #812, opened 2020-05-28
[JS->TS]* Mirgate add_token.js

bh2smith:
This is slightly more involved than usual. Had to migrate a few utility functions required for this script to run. Partial fulfillment of #330.

Note that have adjusted the allowance amount to only approve the minimum amount necessary (i.e. the difference of required and existing allowance).

**Test Plan:**

```sh
yarn build
npx truffle exec build/common/scripts/add_token.js --network rinkeby --tokenAddress 0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea
```


scripts/util.ts
```diff
@@ -0,0 +1,95 @@
...
+import type { BatchExchangeInstance } from "../types/truffle-typings";
+import { factory } from "../src/logging";
+const log = factory.getLogger("scripts.util");
+
+export interface ExchangeToken {
+  id: number;
+  address: string;
+}
+
+export async function getBatchExchange(
```
nlordell, line 11: I think these should probably be removed from their JS counterpart. Note that if they are still being used by other JS scripts, they can `const { getBatchExchange } = require("../build/common/scripts/util")`.

nlordell (approved):
Looks good! I would just like to see the JS version of the ported utility functions removed.

nlordell:
> Note that have adjusted the allowance amount to only approve the minimum amount necessary (i.e. the difference of required and existing allowance).

I think this should be done in a separate PR, Its really hard to see what this change actually is in the diff because all you see is a removed file and an added file.

bh2smith:
>  I would just like to see the JS version of the ported utility functions removed.

Yes, unfortunately there are still other scripts depending on these, but I can try to point them at the new util function.


> I think this should be done in a separate PR,

Cool. Will put it back, this is why I mentioned it...

scripts/add_token.ts
```diff
@@ -0,0 +1,24 @@
+import { addTokens, getBatchExchange, getOwl } from "./util";
+import { factory } from "../src/logging";
+const log = factory.getLogger("scripts.place_order");
```
bh2smith:
```suggestion
const log = factory.getLogger("scripts.place_order");
```
```suggestion
const log = factory.getLogger("scripts.add_token");
```

PR #813, opened 2020-05-28
[JS->TS] Migrate add_token_list.js

bh2smith:
Required a little tweaking to the JSON fetching, but otherwise pretty easy. Partial fulfillment of #330

**Test Plan:**

```sh
npx truffle exec build/common/scripts/add_token_list.js --network rinkeby
```


package.json
```diff
@@ -123,4 +123,4 @@
     "web3-utils": "^1.2.8",
     "yargs": "^15.3.1"
   }
-}
+}
```
nlordell: nit: missing newline at end of file.

package.json
```diff
@@ -74,6 +74,7 @@
   "sideEffects": false,
   "dependencies": {
     "bn.js": "^5.1.2",
+    "cross-fetch": "^3.0.4",
```
nlordell: What was wrong with `node-fetch`? In fact `cross-fetch` just uses `node-fetch` and is intended as an isomorphic library, that is one that works both in the brower and in NodeJS, which is not what we want here since this is used in a Truffle script and not intended to run in the browser.

scripts/add_token_list.ts
```diff
@@ -0,0 +1,50 @@
...
+    const networkId = String(await web3.eth.net.getId());
+    const tokenAddresses = new Set(
+      tokenList.map((token) => {
+        return token.addressByNetwork[networkId] ?? "";
+      }),
+    );
+    // Ensure any non result is removed
+    tokenAddresses.delete("");
+
+    const [account] = await web3.eth.getAccounts();
```
nlordell, line 40: Awesome!

scripts/add_token_list.ts
```diff
@@ -0,0 +1,50 @@
...
+      await fetch(argv.token_list_url)
+    ).json();
+    const networkId = String(await web3.eth.net.getId());
+    const tokenAddresses = new Set(
+      tokenList.map((token) => {
+        return token.addressByNetwork[networkId] ?? "";
+      }),
+    );
+    // Ensure any non result is removed
+    tokenAddresses.delete("");
```
nlordell: nit: I prefer the implementation as it was before, is there any reason this changed?

nlordell (commented):
Just some minor concerns about the new dependency and about changing the implementation from a `Map` to a `Set`.

package.json
nlordell: Maybe you were missing the TS type definitions? `yarn add -D @types/node-fetch`

package.json
bh2smith: Can you please tell me why this keeps happening? I am obviously not intentionally removing this every time.

nlordell: How are you installing dependencies? For me it always adds a new line when installed from the commandline.

package.json
bh2smith: > Maybe you were missing the TS type definitions? yarn add -D @types/node-fetch

Yes this is precisely why I switched, I just thought the easiest thing was to switch over. I can also just include the types if this is the preferred choice here. 

Note, though, that this is **not** a dev-dependency, since it is required for the scripts.

scripts/add_token_list.ts
bh2smith: I was having trouble looping over the object array...

```for (const token in tokenList) { print(token) }``` was only giving me the IDs. After messing with this stupid loop for an hour, I just went with the `Set(map)`

File: package.json (hunk 2 above)
Comment (nlordell) [reply package.json]: > Note, though, that this is *not* a dev-dependency, since it is required for the scripts.

The `scripts/` folder is not included in the npm package (not part of the `package.json` `"files"` setting), and are intended to be run from within the repo, and not from an external project that adds `@gnosis.pm/dex-contracts` as a dependency. This means that, currently, `node-fetch` will never be used as a direct dependency because we don't package any file that uses it in the published npm package. So it should stay as a `devDependency` for now.

If I remember correctly, this was done to keep `@gnosis.pm/dex-contracts` as slim as possible for use with in the front-end (that doesn't actually need these scripts).

If we do want to also publish the `scripts/` directory (which I'm not convinced is a good idea as-is), then this should be done in a separate PR, and then a lot more `devDependencies` need to move to `dependencies` (such as `truffle` for example which is required to run truffle scripts!).

File: scripts/add_token_list.ts (hunk 4 above)
Comment (nlordell) [reply scripts/add_token_list.ts]: `for (const token of tokenList)` might be what you're looking for.

File: package.json (hunk 2 above)
Comment (nlordell) [reply package.json]: > I just thought the easiest thing was to switch over. 

I mean, the only downside is some extra browser specific dependencies get pulled in that will never be used, but in the end its not so important.

It should definitely be a `devDependency` either way though.

Comment (nlordell) [reply package.json]: Turns out `whatwg-fetch` is already a transient dependency, so there is no benefit to using `node-fetch` + `@types/node-fetch` over `cross-fetch`. In fact its slightly worse as there are two package versions to track in the package.json instead of just one. So maybe keep using `cross-fetch`, but move it to be a dev-dependency.

File: scripts/add_token_list.js [hunk 5]
```diff
@@ -1,33 +0,0 @@
...
-  .help(false)
-  .version(false).argv
-
-module.exports = async function (callback) {
-  try {
-    const tokenList = await (await fetch(argv.token_list_url)).json()
-    const networkId = String(await web3.eth.net.getId())
-
-    const tokenAddresses = []
-    for (const token in tokenList) {
```
bh2smith, line 18: @nlordell then this must have been a major typo before.

nlordell: Its because its using
```js
tokenList[token]
```

Which is fine because `token` is, as you mentioned, an ID. So the equivalent, in your debug for loop, would have been:
```ts
for (const token in tokenList) { print(tokenList[token]) }
```

bh2smith: Yea, sure but still subtle and ugly but why would in just arbitrarily choose one of the entries?

bh2smith: In fact, if this list were not sorted by id, or the IDs exceeded the index of the array this would have failed.

bh2smith: Also, notice that this implementation does not consider uniqueness of the token addresses.

package.json
bh2smith: `yarn add`

```suggestion
}

```

fleupold (approved):

scripts/add_token_list.ts
```diff
@@ -0,0 +1,50 @@
...
+  description?: string;
+  rinkeby_faucet?: string;
+}
+
+module.exports = async function (callback: Truffle.ScriptCallback) {
+  try {
+    const tokenList: TokenObject[] = await (
+      await fetch(argv.token_list_url)
+    ).json();
+    const networkId = String(await web3.eth.net.getId());
```
nlordell, line 31: The string conversion surprised me because in the Rust web3 bindings this is a `String` type, but indeed in web3js it is a number. TypeScript FTW!

nlordell (approved):
Looks really good.

scripts/add_token_list.js
nlordell: This is becuase `for (const foo in bar)` iterates for all keys in `bar`. This is used for iterating over keys of an object more than array values. For example:
```js
const bar = { name: "answer", value: 42 };
for (const foo in bar) { console.log(foo) }
// # Prints:
// name
// value
```

This also happens to work for arrays as well:
```js
const bar = ["a", "b", "c"];
for (const foo in bar) { console.log(foo) }
// # Prints:
// 0
// 1
// 2
```

Additionally, it works for sparse arrays:
```js
const bar = [];
bar[1337] = 42;
bar[3] = 3
for (const foo in bar) { console.log(foo) }
// # Prints:
// 3
// 1337
```

This means that the `token` value its iterating on, is actually the index in the array (which happens to also be the token ID).

bh2smith: At least its consistent! Thanks!

PR #819, opened 2020-05-29
[Trivial] Trim whitespace from comma separated script arguments

bh2smith:
Just a little improvement, that might save someone some careful planning someday in the future.

Previously our scripts would only accept arguments like `--tokenIds` with no spaces like `1,2,3`, but now it will handle `1, 2, 3`  just the same.

### DIFF
--- a/scripts/cancel_order.ts
+++ b/scripts/cancel_order.ts
@@ -10,7 +10,7 @@ const argv = require("yargs")
     type: "string",
     describe: "Order IDs to be canceled",
     coerce: (str: string) => {
-      return str.split(",").map((o) => parseInt(o));
+      return str.split(",").map((o) => parseInt(o.trim()));
     },
   })
   .demand(["accountId", "orderIds"])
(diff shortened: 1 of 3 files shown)


fedgiac:
I like the idea, but for this to work the user must surround your arguments in quotes, otherwise the script will use wrong options:
```
$ npx truffle exec build/common/scripts/cancel_order.js --orderIds 1, 2, 3 --accountId 10
> Warning: possible unsupported (undocumented in help) command line option: --orderIds,--accountId
Using network 'development'.

10
[ 1, NaN ]
$ npx truffle exec build/common/scripts/cancel_order.js --orderIds "1, 2, 3" --accountId 10
> Warning: possible unsupported (undocumented in help) command line option: --orderIds,--accountId
Using network 'development'.

10
[ 1, 2, 3 ]
```
It's very tempting to omit the quotes.
What about using [arrays](https://yargs.js.org/docs/#array)?

bh2smith:
Probably won't do.

PR #822, opened 2020-05-29
[JS->TS] Migrate setup_thegraph_data.js

bh2smith:
This is one of the last existing files matching the pattern `scripts/*.js`. Notice that `scripts/.eslintrc` no longer needs to ignore `console.log` in `*.js` becasue there are none and the `utilities.js` file has been deleted (because all used functions from here have been moved into `util.js`.

Partial fulfilment of #330

**Test Plan**

```sh
# T1: Run Ganache with gas limit 8e6
ganache-cli -l 8e6
# T2:  Migrate contracts
truffle migrate
npx truffle exec build/common/scripts/setup_thegraph_data.js
```

Outputs

```
Using network 'development'.

2020-06-02 14:19:11,361 INFO [scripts.setup_thegraph] OWL address 0x6Efe6028802078872651d463803FbDa7F9151542
2020-06-02 14:19:11,400 INFO [scripts.setup_thegraph] Current batchId: 5303667
2020-06-02 14:19:11,449 INFO [scripts.util] Mint 3000000000000000000000 of OWL for user 0x92e6e3Fe05bac3367A43027a6f2d2D38f02836B1
2020-06-02 14:19:11,510 INFO [scripts.util] Mint 3000000000000000000000 of OWL for user 0xBE399be659a80953492E21Be32aAAc1e5C98ba32
2020-06-02 14:19:11,765 INFO [scripts.util] Approving BatchExchange at 0xa3Cc76bf36F75717Be6E3a6F93ca9643169Fac42 for 3000000000000000000000 token 0x6Efe6028802078872651d463803FbDa7F9151542
2020-06-02 14:19:11,812 INFO [scripts.util] Approving BatchExchange at 0xa3Cc76bf36F75717Be6E3a6F93ca9643169Fac42 for 3000000000000000000000 token 0x6Efe6028802078872651d463803FbDa7F9151542
2020-06-02 14:19:11,862 INFO [scripts.util] Approving BatchExchange at 0xa3Cc76bf36F75717Be6E3a6F93ca9643169Fac42 for 3000000000000000000000 token 0x8737F4b313C1Ed19643EBeAcC3621aeFd24E804F
2020-06-02 14:19:11,914 INFO [scripts.util] Approving BatchExchange at 0xa3Cc76bf36F75717Be6E3a6F93ca9643169Fac42 for 3000000000000000000000 token 0x8737F4b313C1Ed19643EBeAcC3621aeFd24E804F
[...]

PR #831, opened 2020-06-02
Type descripency and mint bug

### LINKED ISSUE #830: [Bug] token minting ignored in setup_environment.ts
When writing a for `transitive_orderbook` migration [here](https://github.com/gnosis/dex-contracts/pull/828#issuecomment-637472413) I discovered that only the owl token had been minted to the accounts... After closer look at the following snippet, we see that `token.contract.methods.mint` is not working. This should be `token.mint`.

https://github.com/gnosis/dex-contracts/blob/735c370036971d82b434cb3b09c4ec849ef77d5a/scripts/setup_environment.ts#L45-L57

Furthermore, when attempting to fix this mistake, we realize that we have initialized a list containing two different types:

https://github.com/gnosis/dex-contracts/blob/2025c76ab8a96ff52f30ddbcbf649c4c4b589e1b/scripts/setup_environment.ts#L37-L43

That is, the first entry is `TokenOWLInstance` while all the others are `ERC20Mintable`

bh2smith:
Closes #830 

**Test Plan**

```sh
# T1: Run Ganache with gas limit 8e6
ganache-cli -l 8e6
# T2:  Migrate contracts
truffle migrate
npx truffle exec build/common/scripts/setup_environment.js 
npx truffle exec build/common/scripts/deposit.js --accountId i --tokenId j --amount 10
```

for all `i, j in {0, 1, 2}`.

### DIFF
--- a/scripts/setup_environment.ts
+++ b/scripts/setup_environment.ts
@@ -31,11 +31,11 @@ module.exports = async function (callback: Truffle.ScriptCallback) {
     await owlProxy.approve(instance.address, amount);
 
     log.info(
-      `Beginning environment setup with ${argv.numAccounts} accounts and ${argv.numTokens} tokens`,
+      `Beginning setup with ${argv.numAccounts} accounts and ${argv.numTokens} tokens`,
     );
 
     // Create and register tokens (feeToken is already registered)
-    const tokens = [owlProxy];
+    const tokens = [];
     for (let i = 1; i < argv.numTokens; i++) {
       const token = await ERC20Mintable.new();
       await instance.addToken(token.address);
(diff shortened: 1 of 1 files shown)


scripts/setup_environment.ts
```diff
@@ -44,13 +44,15 @@ module.exports = async function (callback: Truffle.ScriptCallback) {
...
     for (let account = 0; account < argv.numAccounts; account++) {
-      for (let token = 0; token < argv.numTokens; token++) {
-        if (token == 0) {
-          await owlProxy.mintOWL(accounts[account], amount);
-        } else {
-          await tokens[token].contract.methods.mint(accounts[account], amount);
-        }
-        await tokens[token].approve(instance.address, amount, {
+      // Mint and approve OWL
+      await owlProxy.mintOWL(accounts[account], amount);
```
nlordell, line 48: tricky.... Its a different method!

nlordell (approved):
Nice catch!

scripts/setup_environment.ts
bh2smith: It is. I am now going to use my new found understanding here to deal with #822

PR #838, opened 2020-06-05
[Easy] Remove unused dev-dependency (husky)

bh2smith:
This was introduced a while back as an experimental way to easily introduce git hooks. However we haven't been using it anyway, so poof! Its gone!


package.json
```diff
@@ -123,4 +122,4 @@
     "web3-utils": "^1.2.8",
     "yargs": "^15.3.1"
   }
-}
+}
```
nlordell: nit: missing newline.

nlordell (approved):
:boom:

package.json
```suggestion
}

```

PR #861, opened 2020-06-17
bump contracts to version 0.4.0

bh2smith:
It has been over a month since last publish and we have migrated everything in to typescript. This calls for a middle number bump.

Also, I need this version explicitly for the `fetchTokenInfo` function that is now exposed from src


nlordell (approved):
GOGOGO!

PR #1013, opened 2020-10-13
[Easy] Fix broken version bump in solidity-bytes–utils

bh2smith:
This change fixes the broken build introduced by dependabot when attempting to bump our dependency on solidity-byte-utils. That is, a breaking change was introduced on their smart contract we had to make a one line change in `BatchExchange.sol` in order to continue.

Breaking change was introduced here:

https://github.com/GNSPS/solidity-bytes-utils/pull/31

when they changed `toUint` to `toUint256`


nlordell (approved):

nlordell:
I would just merge this directly into `master`. Dependabot will automatically close the PR once it is merged.

PR #1024, opened 2020-10-20
[Easy] update OWL token dependency

bh2smith:
The most recent release of OWL token changed the directory structure of the contracts folder (splitting the different solidity versions). Thus we had to adjust our import statements

### DIFF
--- a/contracts/BatchExchange.sol
+++ b/contracts/BatchExchange.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.5.0;
 import "./EpochTokenLocker.sol";
 import "@gnosis.pm/solidity-data-structures/contracts/libraries/IdToAddressBiMap.sol";
 import "@gnosis.pm/solidity-data-structures/contracts/libraries/IterableAppendOnlySet.sol";
-import "@gnosis.pm/owl-token/contracts/TokenOWL.sol";
+import "@gnosis.pm/owl-token/contracts/5/TokenOWL.sol";
 import "@openzeppelin/contracts/utils/SafeCast.sol";
 import "solidity-bytes-utils/contracts/BytesLib.sol";
 import "./libraries/TokenConservation.sol";
--- a/contracts/DevDependencies.sol
+++ b/contracts/DevDependencies.sol
@@ -5,6 +5,6 @@ pragma solidity ^0.5.0;
 //  contracts during development.
 //
 //  For other environments, only use compiled contracts from the NPM package.
-import "@gnosis.pm/owl-token/contracts/TokenOWLProxy.sol";
+import "@gnosis.pm/owl-token/contracts/5/TokenOWLProxy.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20Detailed.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20Mintable.sol";
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
   },
   "devDependencies": {
     "@gnosis.pm/mock-contract": "^3.0.8",
-    "@gnosis.pm/owl-token": "^3.2.0",
+    "@gnosis.pm/owl-token": "^4.0.0",
     "@gnosis.pm/solidity-data-structures": "1.3.5",
     "@gnosis.pm/util-contracts": "2.0.7",
     "@openzeppelin/contracts": "=2.5.1",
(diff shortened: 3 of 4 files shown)


nlordell:
Does this mean that the contracts need to be re-released?

nlordell (approved):

fleupold (approved):

bh2smith:
> Does this mean that the contracts need to be re-released?

In a sense, yes but I don’t consider it all that urgent. Could wait till next week or even longer. Unless, you think it might make sense to bump the version right here in this PR.

nlordell (changes requested):
So, in checking the NPM module, it looks like we won't need to re-publish a new version after this version bump since it is only a dev-dependency.

However, I think:
- `@gnosis.pm/owl-token@3.2.0` needs to be yanked
- `@gnosis.pm/owl-token@4.0.0` needs to be published.

I think the decision to use a minor version bump instead of a major one was incorrect as the path to the contract has changed, and is in it of itself a breaking change. To illustrate, imagine a package that previously had:

```json
{
  "name": "foo",
  "version": "1.0.0",
  "dependencies": {
    "@gnosis.pm/owl-token": "^3.1.0"
  }
}
```

It would now start resolving to `3.2.0` (since `3.2.0 ∈ ^3.1.0` since `^3.1.0 = [3.1.0, 4.0.0)`). This would cause build errors since the `import` statements would no longer work. This would break any packages depending on `@gnosis.pm/owl-token` that did not use strict versioning (which I imagine is most, as `^version` is the default behaviour for `npm add --save` and `yarn add --save`).

bh2smith:
Yea @nlordell that's some great reasoning. I wish I had come up with this justification when I initially proposed a major version. I hadn't noticed at the time that all the import statements would need to be changed.

I will make the proposed version changes and update here before merging this.

nlordell (approved):
Looks good. Make sure to yank the [previous version](https://www.npmjs.com/package/@gnosis.pm/owl-token/v/3.2.0) as it can still cause unexpected breakage:
```
npm unpublish @gnosis.pm/owl-token@3.2.0
```

Note that (see [here](https://docs.npmjs.com/cli/unpublish)):
> unpublish is only allowed with versions published in the last 72 hours

bh2smith:
> Make sure to yank the previous version

~~Way ahead of you!~~ Wait why is it still there... I definitely removed it this morning...

bh2smith:
@nlordell Look see

```sh
➜  owl-token git:(master) npm unpublish v3.2.0
npm ERR! Refusing to delete entire project.
npm ERR! Run with --force to do this.
npm ERR! 
npm ERR! npm unpublish [<@scope>/]<pkg>@<version>
npm ERR! npm unpublish [<@scope>/]<pkg> --force
➜  owl-token git:(master) npm unpublish v3.2.0 --force
npm WARN using --force I sure hope you know what you are doing.
- v3.2.0
```

bh2smith:
Weird... Now I tried again and seems to be a problem:


```sh
➜  owl-token git:(master) npm unpublish @gnosis.pm/owl-token@3.2.0
This operation requires a one-time password.
Enter OTP: 974223
npm ERR! code E405
npm ERR! 405 Method Not Allowed - PUT https://registry.npmjs.org/@gnosis.pm%2fowl-token/-rev/79-c7ef40a6fb97a85ba07b3284e091b8db - You can no longer unpublish this package.
npm ERR! Failed criteria: 
npm ERR! has dependent packages in the registry
npm ERR! 
npm ERR! Please deprecate it instead:
npm ERR! npm deprecate -f '@gnosis.pm/owl-token@3.2.0' "this package has been deprecated"
npm ERR! To learn more about our unpublish policies, see https://www.npmjs.com/policies/unpublish

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/bensniff/.npm/_logs/2020-10-21T11_58_55_311Z-debug.log
```
Should I deprecate instead? Oh! I see that I should merge this first and remove all other packages depending on this version....

nlordell:
> Should I deprecate instead?

No. The issue with deprecating is that it would still resolve ^3.1.0 to 3.2.0 and break things.

I think you might have to contact NPM support :cry: 

**edit**: Maybe the easiest way to fix this is:
- deprecate v3.2.0
- publish v3.2.1 that is exactly the same as v3.1.0

Then we don't have to contact support any more.

I'm going to do that right now.

**edit2**: all done.

bh2smith:
I found it, this project had depended on it. https://github.com/gnosis/dex-liquidity-provision/pull/547

